PR #15, opened 2019-02-21
Initial dict_list function

Datseris:
Currently this function "fails" because for example if a key of the dictionary `d` is a string then the function expands over it and makes the entries `Char`:

```
julia> d = Dict(:a => [1, 2, 3], :b => 4, :c => "test")
Dict{Symbol,Any} with 3 entries:
  :a => [1, 2, 3]
  :b => 4
  :c => "test"

julia> dict_list(d)
12-element Array{Dict{Symbol,Any},1}:
 Dict(:a=>1,:b=>4,:c=>'t')
 Dict(:a=>2,:b=>4,:c=>'t')
 Dict(:a=>3,:b=>4,:c=>'t')
 Dict(:a=>1,:b=>4,:c=>'e')
 Dict(:a=>2,:b=>4,:c=>'e')
 Dict(:a=>3,:b=>4,:c=>'e')
 Dict(:a=>1,:b=>4,:c=>'s')
 Dict(:a=>2,:b=>4,:c=>'s')
 Dict(:a=>3,:b=>4,:c=>'s')
 Dict(:a=>1,:b=>4,:c=>'t')
 Dict(:a=>2,:b=>4,:c=>'t')
 Dict(:a=>3,:b=>4,:c=>'t')
```

In general it is difficult to predict whether any arbitrary used-defined type should be iterable or not in this case. However establishing the rule that only top level `Vector` types are expanded would make this general and intuitive at the same time.

The problem is, I don't know how to do it. Any help appreciated.

### DIFF
--- a/docs/src/project.md
+++ b/docs/src/project.md
@@ -89,7 +89,7 @@ BSON.bson(datadir()*"sims/test.bson", file)
 ```
 
 ## Reproducibility
-This project setup approach that DrWatson suggests has a very big side-benefit: it is fully reproducible firstly because it uses Julia's suggested project structure and secondly because the navigation only uses local directories.
+This project setup approach that DrWatson suggests has a very big side-benefit: it is fully reproducible firstly because it uses Julia's suggested project structure, secondly because the navigation only uses local directories and lastly because it is a Git repository.
 
 If you send your entire project folder to a colleague, they only need to do:
 ```julia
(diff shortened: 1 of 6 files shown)


Datseris:
I've also added and exported two more functions in this: `current_commit, tag`. Making the same version of `dict_list` for named tuples was too much for me right now. I'll merge when tests pass.

PR #19, opened 2019-02-27
Result collection

### LINKED ISSUE #3: function addrun!
A function `addrun!(csvfile, d)` that adds a simulation/experiment/whatever run to a csv file would be great.

This will allow to make csv tables out of existing simulations. Here is some things that quickly come to mind:

1. The function should operate similarly to `savename`, i.e. allow choosing which keys to save, which types, etc.
2. if the `csvfile` doesn't exist, it creates a new one and adds headers.
3. This function should be able to cooperate with (or even internally create) `DataFrame`.
4. It should somehow allow adding a run that has some _additional_ parameters. How to do this is to be discussed. It could add `missing` entries to all other existing runs that do not have this parameter.
5. As above, when adding a run with _fewer_ parameters, it should add `missing`/NA/Idon'tknowwhat to the parameters that do not exist.

JonasIsensee:
Closes #3

I cleaned up my implementation for results collection and wrote a docstring.

I'm still not sure how to properly test it.
It'd probably best to write a separate script thats creates a nice file structure with dummy results
with lots of pitfalls and then call this function on it.


Datseris:
Awesome thanks! I'll have to review this tomorrow. 

I read the docstring, but I do have one question. This seems to already be a "higher level" functionality that takes place _after_ all files are saved.

Do you think it will be easy for us to also produce the function `addrun!` from this, which takes as in input the result dictionary and adds it to a dataframe directly? So that you don't call this `collect_results` after your script is over, but at the end of your run, something like:


instead of:
```
for parameters
  # lalala
  dict = # has all my results for this run.
  save(dict, savename with parameters)
end
collect_results()
```

to be also possible to do:
```
for parameters
  # lalala
  dict = # lalala
  addrun!(file_with_dataframe, dict
end
```

JonasIsensee:
Yes and no.
of course you could call this function at the end of every run but that wouldn't be safe for the same reason as discussed in #3.
I run 30 jobs in parallel and would not want to risk them colliding when writing to the same file at the same time.

Since I don't know any decent fix to this I avoided it.
I typically call this function manually before downloading the results file.
that way I still have all the newest results with just the tiniest amount of inconvenience.

JonasIsensee:
A somewhat hacky solution would be to always keep a separate Julia session running that does something like
```
while true
collect_results()
sleep(1000)
end
```

You could even embedd this into the `multiprocessing.jl` file with an `@async` task

Datseris:
I am reviewing this now. I see an immediate problem: how does this code load `jld2` files? It is using `BSON` but `BSON` cannot load `jld2` files. I guess we need to also put FileIO as a dependency? but this is too hard of a dependency. :(

src/result_collection.jl
```diff
@@ -0,0 +1,92 @@
+using DataFrames
+using BSON
+
+function merge_df(df1, df2)
+    if names(df1) == names(df2)
+        return vcat(df1, df2)
+    else
+        for m ∈ setdiff(names(df1), names(df2))
+            df2[m] = [missing]
```
Datseris: Here why does the `missing` entry need to be in a vector?

src/result_collection.jl
```diff
@@ -0,0 +1,109 @@
...
+    data_folder = joinpath(datadir(), "results"),
+    filename = joinpath(datadir(),"results_dataframe.bson"),
+    valid_filetypes = [".bson", ".jld2", ".jld"],
+    kwargs...)
+
+    df = isfile(filename) ? BSON.load(filename)[:df] : DataFrame()
+
+    for (root, dirs, files) in walkdir(data_folder)
+        for file in files
+            is_valid_file(file, valid_filetypes) && continue
```
Datseris: This is wrong, right? Either do `!is_valid...` or use `||` instead of `&&`

src/result_collection.jl
```diff
@@ -0,0 +1,109 @@
...
+end
+
+is_valid_file(file, valid_filetypes) =
+    any(endswith.(Ref(file),valid_filetypes))
+
+function to_data_row(data,
+        white_list = collect(keys(data)),
+        black_list = [],
+        special_list = [])
+    cnames = setdiff!(white_list, black_list)
```
Datseris: I think this kind of `setdiff!` processing should happen once in the main function instead of being repeated everytime in `to_data_row`.

Datseris: Ah now I see. this cannot be done, because you are taking into acount already the possibility of different datasets with different keys each. Very good!

src/result_collection.jl
JonasIsensee: I'm adding a _column_ to a `DataFrame`.
This is a column with just one single row but it need to be a column-vector nonetheless

src/result_collection.jl
JonasIsensee: yes, this is a typo.

JonasIsensee:
This function is _designed_ to load the data by itself.
If you do not want for `FileIO` & `JLD2` to be dependencies,
please provide a different approach.

Datseris:
I'll ask on discourse how to handle such case.

Datseris:
@JonasIsensee have you seen this: https://github.com/MikeInnes/BSON.jl/issues/25 ?

JonasIsensee:
> @JonasIsensee have you seen this: [MikeInnes/BSON.jl#25](https://github.com/MikeInnes/BSON.jl/issues/25) ?

Now I have..

Datseris:
Someone said that "BSON has a limit around at 2GB for saving files". After that it fails.

Datseris:
@JonasIsensee I think it is easy to write this function to use something specific to read and write. We should not try to use multiple saving modules, because this is very hard to get consistent results. 

I know how to do this, but we still need a test case. Once you've added some tests here I'll test this with the existing packages that can read/write.

If they all fail, we will have to look at `Serialize`.

JonasIsensee:
Definitely not a fan of all of this.
This package is meant to help with science and in particular with reproducibility.
We really need a _stable_ and _maintaned_ format for saving our data.
(this mainly concerns the job-result-files)
For the result-dataframe you can always just rerun the script if necessary.

If necessary we may need to consider other formats non-native to julia.

Datseris:
>  We really need a stable and maintaned format for saving our data.

But the point is that there isn't any Julia package that does this. There isn't any Julia code that does this in general, since `serialize` depends on the used Julia version.

Datseris:
> Definitely not a fan of all of this.

Well it's simple math really. The goal is to have a reproducible framework with as less bugs as possible, while still being as robust and consistent as possible.

Is it easier if:

1. You use 3 totally different saving packages, each with no relation whatsoever with each other.
2. You use 1 package.

JonasIsensee:
> > We really need a stable and maintaned format for saving our data.
> 
> But the point is that there isn't any Julia package that does this. There isn't any Julia code that does this in general, since `serialize` depends on the used Julia version.

This is not quite true.
`JLD.jl` does not use julia's `serialize` and is compatible across multiple versions.
(but of course has other drawbacks)

Another thing to consider is the size of serialized data.
I noticed that in the case of the `results_dataframe` the `.jld2` was just half the size as compared
to the `.bson` file.
This may be relevant when deciding on which saving format `DrWatson.jl` endorses.

Datseris:
> Another thing to consider is the size of serialized data.

Yes this is indeed important. Another important factor is the speed of saving, although by far the least important.

What I believe is the _most_ important factor is the dependency tree and how possible and easy it is to install a given package on a given computer. I know for sure it was near impossible for me to use JLD on the old clusters we had, simply due to the huge amount of external dependencies. JLD2 was much better and seemed to be installed without problems.

What I've noticed so far though is that BSON has only Julia in its dependencies: https://github.com/MikeInnes/BSON.jl which raises no doubts whether it is installable anywhere that Julia is installable. 

Another thing to consider, although not very important is compilation speeds. BSON pre-compiles and compiles on `using` almost instantly. I was very impressed with this.

---

@JonasIsensee You have a better memory of what problems you had when using JLD2, and why you switched to BSON. I remember that there was problem with `missing` data for JLD2. But what problems did you have with BSON? Can you write them here for achiving?

JonasIsensee:
I have not had any problems with `JLD2` except for what you already mentioned.
It works fine for saving the results from jobs,
and only fails to properly save things that include `missing` values.

BSON can correctly handle `missing` data. I have not had any problems with BSON so far.
(only that the files seem to be larger than the corresponding `.jld2` files)
and you mentioned that it may not be able to work with very large >2GB files.

JonasIsensee:
Hey,
I finally got around to writing script for you that generate a file structure
you can call `collect_results` on.
(the function also works now)
There still is the major issue with dependencies, so you will have to load the packages into your main scope to test all this.

Datseris:
Hi, @JonasIsensee , I appreciate your work on this a lot.

Since today I took a day off I spent some time here. All in all this is great, but I did do some major changes in my latest commit:

* Removed FileIO as a dependency. Also, `load` was ambiguous as it is exported by both `BSON`, `FileIO`. If someone finally puts `BSON` to `FIleIO` then we make the dependency `FIleIO` instead and never have to worry about loading.
* Due to the above the valid filenames are now only `.bson`. Really hope we can use FIleIO interface in the future...
* I have most default values for collect_dataframes. For me it makes no sense to automatically read subfolders. There is now an option to either read only the given dir or also all subfolders.
* To be more in line with the above, I have changed the default saving name.
* Added option to not save `df` at all.
* Added warning for `:path`.
* Fixed the not checking for valid filenames.
* Updated tests to work for me
* Added some actual tests

For most scientists, most subfolders have different kind of results, and it doesn't make sense to mix them into the same dataframe.

---

Have a look and let me know. If there isn't any further discussion we can merge this and we can also look for Hooking up BSON to FileIO. THis will solve all problems since users can just _choose_ which file endings are correct and FIleIO takes care of everything else!

Datseris:
Oh, we should add more tests to this I feel. Seems like it is a complicated functionality and the tests are super basic only...

Datseris:
@JonasIsensee I have two comments. When running the testsuite I get : 
```
┌ Warning: KeyError("c")
└ @ DrWatson C:\Users\datseris\.julia\dev\DrWatson\src\result_collection.jl:49
┌ Warning: KeyError("c")
└ @ DrWatson C:\Users\datseris\.julia\dev\DrWatson\src\result_collection.jl:49
```
I don't know where this comes from (especially since `"c"` is blacklisted).

Maybe we should consider how to improve these wornings, e.g. pass the filename to the function as well so that it warns and lists the file.

---

The second comment is what you said:

> EDIT: BSON does have one major flaw. ATM It can only reconstruct types defined in Main.

Yeah unfortunately `collect_results` doesn't work with BSON if one wants to load the DataFrame.... maaaan what the hell is this situation!!! On the other side we can't use JLD2 because it corrupts any array with `Missing` union types!!!

---

Here is an idea: We put `collect_results` behind a `@require` code block. This way the function is accessible only when the user is doing `using DataFrames`. This also drops the `DataFrames` dependency which is cool!

JonasIsensee:
I don't know about the warnings - I'll have to take a look at that.

The `@require` solution sounds good to me.
Also, I'm working on a PR to make FileIO compatible with BSON right now.
maybe we'll have that in a few days.

Otherwise looks good to me.
We can't use it in `EmbeddingResearch` since all result files are `.jld2` and I really don't want to change this in the middle of a project.

Datseris:
> We can't use it in EmbeddingResearch since all result files are .jld2 and I really don't want to change this in the middle of a project.

Yeah don't do that use either the local version on the embedding repo or do a branch in this branch 2 commits in earlier.

But if we get the BSON FileIO thingy, then you won't have to worry about it and in the end be able to use everything!

---

Alright, then I'll put it behind a `require` for now and we see in the future if BSON has fixed the problem.

Datseris:
Alright @JonasIsensee I have resolved the obscure warning. There wasn't something wrong, just some dictionaries not having the key required by the functions. 

I've improved the warning as well.

Issue #28, opened 2019-03-27
Recollection option for collect_results


Datseris:
After a bit of usage of `collect_results` it seems like an option to "recollect" is useful.

By that I mean that by giving a keyword argument e.g. `recollect = true` all _existing_ data should be re-collected instead of being skipped.

I believe that this functionality should _not_ delete the saved dataframe and just start over. The reason for that is some old results still saved in the dataframe may not exist in the data folders any more. Then it would be irreversible to remove them. The best approach would be to just "replace" existing files if they still exist.

@JonasIsensee do you have any idea how this can be done? If some please give some help here. (Or if you can even implement this then you're the boss!)

JonasIsensee:
instead (or additionally) to another argument we could consider exporting an _inplace_ and _outofplace_
version marked by `!`.
See my PR for an implementation for _backups_.
Feel free to integrate that into the relevant saving functions.

Datseris:
This is a wonderful idea, i really like it!!!

I think it is clear that the out-of-place version will always produce a new dataframe with everything re-processed. (i.e. as if running the current version but first deleting the existing saved dataframe). We can also implement a `safe` keyword after #29. 

Now about the in place version, how do you envision it? From my perspective it always adds new result-files in place in the given directory by loading it (i.e. exactly what the current function does). However for the files already processed there are two approaches:

1. Re-process them.
2. Don't reprocess them (i.e. what the current function does).

Maybe we can keep both approaches through a keyword? i.e. `reprocess = false` by default?

Datseris:
The `reprocess = true` keyword is actually equivalent with deleting the dataframe file and thus it is unnecessary.

Issue #2, opened 2019-01-29
Automated generation of project based on a text file


Datseris:
In my dreams, the function `initialize_project` somehow reads the project specification text file and then knows how to create the project.

This is great for documentation and consistency: the project specification text file `src/project_structure` is loaded and expanded into the documentation. The same file is used for project creation and thus everything is robust and consistent.

The question is, how to structure the project_structure file and how to read it in Julia to make this possible?

taqtiqa-mark:
Thanks for putting effort into making Dr Watson.  It looks promising as a tool to allow various domains structure their projects in a way that best suits that domain.

Hope this helps.....

The current default is described in `src/defaults/project_structure.txt`
This could be:
1. Formatted as some TOML file
1. If there is:
   - no `some-valid-drwatson-layout.toml` given at the command line
   - no `.drwatson.toml` in the `~/.drwatson` or `~/` user-space folders
1. Parsed and used to create the folders
1. Saved in the project as `.drwatson.toml`

The current configuration would look like this:
```
#
# Note each entry in `project:dirs` must be a sub-table that belongs to 
# the `project` element above it
#
[[project]]
dirs = [  "data"
              "plots",
              "notebooks",
              "papers",
              "scripts",
              "src",
              "test"]
    files = ["README.md",
                "Manifest.toml",
                "Project.toml" ]
    [[project.data]]
     dirs = [ "sims",
                  "exp_pro",
                  "exp_raw"
     ]
    [[project.plots]]
     dirs = [ ""]
    [[project.notebooks]]
     dirs = [ ""]
    [[project.papers]]
     dirs = [ ""]
    [[project.scripts]]
     dirs = [ "" ]
     files = ["intro.jl"]
    [[project.src]]
     dirs = [ ""]
    [[project.test]]
     dirs = [ ""]
     files = ["runtests.jl"]
```

This way each domain could setup projects as best sits them.

Datseris:
The suggestion is quite nice, since it allows for customization.

But one has to weight the cost carefully versus the added complexity. This TOML-based setup is significantly more complex than the current file `src/defaults/project_structure.txt`. As I have no knowledge of TOML or parsing TOML, is it possible to somehow transform this file into something a bit more familiar and a bit more tree like, like the `src/defaults/project_structure.txt`?

asinghvi17:
[...]

Issue #1, opened 2019-01-29
Perfect project structure


Datseris:
What is the perfect project structure? The project structure is described in the `src/project_structure` file. I think the original version is already quite okay. Keep in mind we want to keep things basic: adding more folders, not using other folders, all should be flexible.

Datseris:
The file is in src/defaults/project_structure.txt

PR #29, opened 2019-03-28
WIP: backup existing files instead of overwriting them

JonasIsensee:
This addresses #28 

This is basic functionality without integration into the other algorithms.

The idea is inspired by GROMACS.
If the `test.bson` to be saved to already exists,
the old file will be moved to `#1#_test.bson`.
If this one exists already as well, the file previously called `#1#_test.bson` will be renamed to `#2#_test.bson`.

The newest file is always called `test.bson` oldest file is called `#42#_test.bson` (if there are 43 saved files in total).

What do you think?


Datseris:
Hi Jonas! This is really nice! I really like this and will definitely review "properly" and merge once this has tests implemented! You might want to write a basic doc-string as well. I can improve it if need be.

I assume you have checked and indeed one can use `#` in file names in all major operating systems, right?

I am not sure about how this:
```
run(`mv $cur_path $new_path`)
```
Why do you have to use the command line in the first place? You can use pure Julia code from the FileSystem module here: https://docs.julialang.org/en/latest/base/file/#Base.Filesystem.mv the funciton `mv` exists already. Please consider doing that; it will save us a lot of headaches later on.

We can see about integrating this elsewhere as well, but let's put in some basic tests and docs first. I assume having a keyword `safe = true` for e.g. `tagsave` would be nice from an API perspective.

---

I am confused about this addressing #28. Maybe you got confused. #28 is only for re-processing the data that are already existing in the collection dataframe. It is not about saving the file of the dataframe itself. For example let's say I have already processed files `a` and `b` (and thus these files are in the `:path` column) but I want to _reprocess_ them so I can use e.g. the `special_list` argument.

Datseris:
I would also change the main function name from `save_safely` to `safesave`

Datseris:
Another thing that I believe can be improved: the position of the numbering scheme. Currently it is always in the start, but this makes for a bad displaying in the containing folder. Since most OS sort files alphabetically the "new (old)" files with `#` will be far away from the new new files. 

Isn't it a better approach to put the number scheme before the file ending? E.g. before `.bson`. So `test.bson` will become `test_#1#.bson` ? In addition, is there a reason for the second `#`? Isn't `test_#1.bson` enough due to Regex? Because if it is enough we should do it like this; the less characters in a file's name the better it is.

p.s.: the file separator position can be found by:
```julia
a = "test.ka.bson"
s = findall(isequal('.'), a)[end]
```

JonasIsensee:
Indexing into strings is BAD. I found a different solution using regexp though.
Let's just hope, no one wants to use a file type that has two suffixes.
Wouldn't want any `.tar.gz` be renamed to `.tar#1.gz` ....

All common file systems seem to support # in filenames.

JonasIsensee:
Uh... a rebase seems to really spam github.
Sorry about that.

I'm done here.
There are two merge conflicts here that I can't seem to resolve.
Github claims that my code conflicts with an empty line on master
but I can't reproduce that locally.

Datseris:
@JonasIsensee Something happened on GitHub that I cannot comprehend. GitHub initially told me that you have created _new_ files for your new code. So in my conflict resolution I removed conflicting parts. But now the code is gone. Whaaat!!!

Also since there seem too many unnecessary commits. Did you try to rebase into your local master, which was unsynced for a month?

In any case, the easiest way forward is to close this and simply open a new PR. Can you please do that? (By the way you have push access to DrWatson.jl).

PR #41, opened 2019-04-25
Improve typing on return type of `dict_list`

jondeuce:
I came across the following minor incovenience:
```julia
julia> d = Dict(:x => [1], :y => [2.0])
Dict{Symbol,Array{T,1} where T} with 2 entries:
  :y => [2.0]
  :x => [1]

julia> dict_list(d)
1-element Array{Dict{Any,Any},1}:
 Dict(:y=>2.0,:x=>1)
```
The return type of `dict_list` is a `Vector` of `Dict{Any,Any}`, when the type of the key (at least) should be able to be inferred. It turns out that it is due to merging the `iterable_dict` with the `non_iterable_dict` inside the `dict_list` method, where `non_iterable_dict` is an empty `Dict{Any,Any}`.

The fix is simply to check for emptiness, and only merging if they are both non-empty. Then, we get the expected output:
```julia
julia> dict_list(d)
1-element Array{Dict{Symbol,Real},1}:
 Dict(:y=>2.0,:x=>1)
```
I've added some corresponding tests, too.

### DIFF
--- a/src/saving_tools.jl
+++ b/src/saving_tools.jl
@@ -174,7 +174,13 @@ function dict_list(c::Dict)
     vec(
         map(Iterators.product(values(iterable_dict)...)) do vals
             dd = Dict(keys(iterable_dict) .=> vals)
-            merge(non_iterable_dict, dd)
+            if isempty(non_iterable_dict)
+                dd
+            elseif isempty(iterable_dict)
+                non_iterable_dict
+            else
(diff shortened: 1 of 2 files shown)


Datseris:
Excellent! Code is fully LGTM! Thanks for your contribution and welcome to the contributor list! :)

jondeuce:
No problem :) Thanks for your great work on this package.

Issue #40, opened 2019-04-25
Default project should include test directory


jessebett:
Encourage scientists to write unit test for `src`!

Datseris:
To implement this one has to : 
1. Add the path here: https://github.com/JuliaDynamics/DrWatson.jl/blob/master/src/project_setup.jl#L118
2. Document it in this text file: https://github.com/JuliaDynamics/DrWatson.jl/blob/master/src/defaults/project_structure.txt

I would suggest also making a default `runtests.jl` file.

This is a quite easy issue, perhaps @jessebett its your chance to become a contributor? ;)

PR #43, opened 2019-04-28
Add test directory to default project

jessebett:
Addressing #40 ;)

Now that I think about it, I'm not sure if `testdir()` will ever actually be useful...


Datseris:
Perfect, welcome to the contributor list!

Yeah I also don't think `testdir()` will ever be used, but since it is not an exported name I am fine with it.

PR #64, opened 2019-07-14
Extend path accessing functions to accept arguments

### LINKED ISSUE #47: Convenience signature datadir(args...)
I propose a single one-liner 
`DrWatson.datadir(arg1, args...) = joinpath(datadir(), arg1, args...)`

This will allow the following common pattern to be shorter.
```
path = joinpath(datadir(), "results", "mymodel")
path = datadir("results", "mymodel")
```

The following 
`datadir(args...) = joinpath(datadir(), args...)`
causes a stackoverflow because `args...` may also be no arguments at all.

tamasgal:
This implements methods for the path accessing functions which accepts subdirectories as arguments, as initially proposed by @JonasIsensee in https://github.com/JuliaDynamics/DrWatson.jl/issues/47

Closes #47

### DIFF
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -1,9 +1,11 @@
 ##########################################################################################
 # Project directory
 ##########################################################################################
-export projectdir, datadir, srcdir, plotsdir, scriptsdir, papersdir, testdir
+export projectdir, datadir, srcdir, plotsdir, scriptsdir, papersdir
 export projectname
 export findproject, quickactivate
+
+@deprecate scriptdir scriptsdir #TODO: remove in next release
 
 """
     projectdir()
@@ -12,25 +14,21 @@ Return the directory of the currently active project.
 ```julia
 projectdir(args...) = joinpath(projectdir(), args...)
 ```
-Return the directory of the `folder` in the active project.
+Join the path of the currently active project with `args`
+(typically other subfolders).
 """
 projectdir() = dirname(Base.active_project())
 projectdir(args...) = joinpath(projectdir(), args...)
 
 
 # Generate functions to access the path of default subdirectories.
-for dir_type ∈ ("data", "src", "plots", "scripts", "papers", "test")
+for dir_type ∈ ("data", "src", "plots", "scripts", "papers")
     function_name = Symbol(dir_type * "dir")
     @eval begin
         $function_name() = projectdir($dir_type)
         $function_name(args...) = projectdir($dir_type, args...)
     end
 end
-
-# backward compatibility
-scriptdir = scriptsdir
-export scriptdir
-
 
 """
     projectname()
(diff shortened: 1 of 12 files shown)


Datseris:
Absolutely perfect!!! A quick comment: should we write some really basic tests for this, simply for long term safety?

Datseris:
Also, the change should be reflected in the docstrings and the changelog (I can do this as well)

tamasgal:
Yep, I apologise for not providing unit tests, I’ll write a few later!

tamasgal:
I also realised that `scriptsdir` is actually `scriptdir`, so that was a breaking change. I think using plurar makes more sense w.r.t. to uniformity?

I also added some tests, PR will follow in a few seconds.

tamasgal:
The windows tests due to the `\` path separator failed. Maybe it's a good time to review the the code and avoid any hardcoded usage of the `\` things. What do you think? This would mean replacing all the `*` stuff with `joinpath()`, at least in the code base. The user can of course do what they want ;)

tamasgal:
I realise that the path separator issues goes much deeper than I thought... Not sure how to proceed since the clean solution will break backward compatibility for users who were used to do the `*`-joining... 😕

Datseris:
@tamasgal Don't worry about backwards compatibility. We are not 1.0 yet exactly for this reason, because I am well aware that DrWatson needs improvements still. Let's simply try to give warnings in the most crucial aspects, but otherwise don't stress that much.

> I think using plurar makes more sense w.r.t. to uniformity? 

Yes I agree.

tamasgal:
Ah OK, still I am always very careful about breaking existing API 😆 

I started to clean things up, but it still needs improvement. I also just realised that there is some implicit functionality in `savename`, when the target ends with the path separator it changes the behaviour slightly. Not sure how to handle this in a nice way, currently I defined `PATH_SEPARATOR` and check for equality, since Julia is not providing public access to that value (although it's available in the private namespace inside `Base`).

src/project_setup.jl
```diff
@@ -1,28 +1,34 @@
...
+
 
-datadir() = projectdir()*"data/"
-srcdir() = projectdir()*"src/"
-plotsdir() = projectdir()*"plots/"
-scriptdir() = projectdir()*"scripts/"
-papersdir() = projectdir()*"papers/"
-testdir() = projectdir()*"test/"
+# Generate functions to access the path of default subdirectories.
+for dir_type ∈ ("data", "src", "plots", "scripts", "papers")
```
Datseris: hell yeah, metaprogramming for the win!!!

tamasgal:
Haha thanks! I am really happy to have found such a nice project. However, I almost feel bad that I am kicking out the door with such a seemingly tiny change (let alone I never use/used an OS with a different path separator than `/` :joy: )

Datseris:
> seemingly tiny change 

Heh, it is this tiny inaccuracies that can become a tornado of problems later on! Imagine if you haven't done this and we had 1.0 out. And suddenly this becomes a problem. BOOOOOOOOOM

Issue #38, opened 2019-04-18
reverse engineer `savename`


Datseris:
It would be nice to have a function that attempts to reverse engineer a string that is the output of `savename`. This function would always return `prefix, params, suffix` with `prefix, suffix` being strings (if non found then the empty strings) and `params` always a Dict with String keys (arbitrary choice).

To make this function is easy: first split by the connector argument (by default `_`), then identify `prefix, suffix` if any, and then find all the pairs connected by `=` and make them into the Dict.

Datseris:
thanks to @sebastianpech

PR #65, opened 2019-07-14
Add function for reverse engineering savename

sebastianpech:
Implements a function to parse the savename into a dict and extract the pre- and suffix (#38).
I thought of some special cases
- prefix is a path containing a folder named something like "a=1.23_mode=double"
- key or value contains `connector` (no precise solution, I'm adding them to the value part)
- last number is a float so the "." doesn't denote the start of the suffix

### DIFF
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -270,8 +270,7 @@ function parse_savename(filename::AbstractString;
     # Prefix can be anything, so it might also contain a folder which's
     # name was generated using savename. Therefore first the path is split
     # into folders and filename.
-    savename_split = splitpath(filename)
-    prefix_part, savename_part = savename_split[1:end-1],savename_split[end]
+    prefix_part, savename_part = dirname(filename),basename(filename)
     # Extract the suffix. A suffix is identified by searching for the last "."
     # after the last "=".
     last_eq = findlast("=",savename_part)
@@ -301,6 +300,8 @@ function parse_savename(filename::AbstractString;
         # Of course the connector symbol could also be part of the variable name.
         prefix, _parameters = name[1:first(first_connector)-1], name[first(first_connector)+1:end]
     end
+    # Add leading directory back to prefix
+    prefix = joinpath(prefix_part,prefix)
     parameters = Dict{String,Any}()
     # Regex that matches smalles possible range between $connector and =.
     # This way it is possible to corretly match something like
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -1,4 +1,4 @@
-export savename, @dict, @ntuple, @strdict
+export savename, @dict, @ntuple, @strdict, parse_savename
 export ntuple2dict, dict2ntuple
 
 """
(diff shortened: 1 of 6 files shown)


test/naming_tests.jl
```diff
@@ -57,3 +57,75 @@ s = savename(e; allowedtypes = (Any,), expand = ["c"])
 @test ')' ∈ s
 @test occursin("a=3", s)
 @test occursin("b=4", s)
+
+# Tests for parse_savename
+
+function test_convert(prefix::AbstractString,c;kwargs...)
+    name = savename(prefix,c;kwargs...)
+    prefix, _b, suffix = DrWatson.parse_savename(name)
+    dicts_equal(_b,c) && prefix == prefix && suffix == ""
```
Datseris: Hm, maybe `prefix == prefix ` here is a typo? (This can only be `true`, I would assume that you wanted to name the prefix of the previous line as something like `prefix2`

test/naming_tests.jl
```diff
@@ -57,3 +57,75 @@ s = savename(e; allowedtypes = (Any,), expand = ["c"])
...
+end
+
+function test_convert(c,suffix::AbstractString;kwargs...)
+    name = savename(c,suffix;kwargs...)
+    prefix, _b, suffix = DrWatson.parse_savename(name)
+    dicts_equal(_b,c) && suffix == suffix && prefix == ""
```
Datseris: same comment here for `suffix`.

Maybe I have misunderstood something, so please correct me if need be.

src/naming.jl
```diff
@@ -250,3 +250,89 @@ end
 function dict2ntuple(dict::Dict{Symbol, T}) where T
     NamedTuple{Tuple(keys(dict))}(values(dict))
 end
+
+
+"""
+    parse_savename(filename::AbstractString; kwargs...)
+                        parsetypes = (Int, Float64),
+                        connector = "_")
```
Datseris: I assume that the extra keywords here were forgotten in the call signature, since you very nicely explain them later on in their own section.

src/naming.jl
```diff
@@ -250,3 +250,89 @@ end
...
+    prefix_part, savename_part = dirname(filename),basename(filename)
+    # Extract the suffix. A suffix is identified by searching for the last "."
+    # after the last "=".
+    last_eq = findlast("=",savename_part)
+    last_dot = findlast(".",savename_part)
+    if last_dot == nothing || last_eq > last_dot
+        # if no dot is after the last "="
+        # there is no suffix
+        name, suffix = savename_part,""
+    else
```
Datseris: Wow, this is very impressive; you have implemented a very smart function, congrats! :)

src/naming.jl
```diff
@@ -250,3 +250,89 @@ end
...
+        prefix, _parameters = name[1:first(first_connector)-1], name[first(first_connector)+1:end]
+    end
+    # Add leading directory back to prefix
+    prefix = joinpath(prefix_part,prefix)
+    parameters = Dict{String,Any}()
+    # Regex that matches smalles possible range between $connector and =.
+    # This way it is possible to corretly match something like
+    # string_with_underscore_a=10.1 as the next match for the regex is
+    # string_with[_underscore_a=]10.1
+    name_seperator = Regex("$connector[^$connector]+=")
```
Datseris: Again, this is excellent; I would never have been able to do this myself, since I don't really know much about Regex... Jonas knows much more, I assume he would be even happier than me to see this!

test/naming_tests.jl
sebastianpech: Oh yes, that's nonsense :) I'll fix it

src/naming.jl
sebastianpech: Oh yes. I usually have them in the call signature, but wanted to match your style

sebastianpech:
Thanks for the nice package. I've attempted to implement something similar many times in the recent year, but was never fully satisfied.
DrWatson doesn't yet cover all the stuff I need (or use) in my workflow but I'm sure we manage to create a  nice package as a community.

sebastianpech:
I changed my mind about parsing the connector symbol. In my opinion it's more likely (especially for the underscore) to have it in the variable name instead of the variable value. So I turned the regex around and now the value is delimited by = and the closest connector symbol. So stuff like `l_ac=10.1` is now correctly parsed but `some_string=string_with_underscore_l_ac=10.1` results in `some_string = "string"` and `with_underscore_l_ac=10.1`. I also added a note to the docstring about the parsing behavior.

sebastianpech:
Should I merge the current master here?

JonasIsensee:
Hey, good work.

Just a quick comment.
if you really do want to allow for values with _
You could probably still escape them in the savename
function.
`some_string='my_string'_next_var=5`

But I'm not sure sure if that this would be a good idea..
(im also not sure if those are valid characters for filenames in Windows)

sebastianpech:
Thanks! Yes escaping would be an option, but as you said, I also don't think it's a very good idea to put `'` into the filename or actually in general to put too much data into the filename.

sebastianpech:
One thing just came to my mind on the way to work. I'm not taking into account that `connector` can be an arbitrary string. Do you want to keep allowing that or should we restrict it to beeing a `Char`?

Datseris:
> Should I merge the current master here?

Yes please.

> I'm not taking into account that connector can be an arbitrary string. Do you want to keep allowing that or should we restrict it to being a Char?

I don't think this is worth the effort to expand to multi-string connectors. In my eyes `parse_savename` should target the basic outcomes of `savename`, not complicated ones. I find it fair to leave it as is, and you can throw an error in multi-string connectors.

> You could probably still escape them in the savename function.
> `some_string='my_string'_next_var=5`
> But I'm not sure sure if that this would be a good idea..

I agree, I don't think this is a good plan either.

PR #69, opened 2019-07-16
Use git describe if possible for tagging

tamasgal:
I use annotated Git tags (`git tag -a v0.1.2`) to mark the stage of the developments even in my research repositories and then `git describe` gives a nice human readable representation of the HEAD. I think it would be nice if this was used in `tagsave` too.

For example if `git describe` spits out this: `12.0.0-alpha.1-8-g6f60bd1d-D`, I immediately know that it's 8 commits ahead of version 12.0.0-alpha.1, it's a dirty repository (`-D`) and the last commit is `6f60bd1d`.

My PR just falls back to the commit hash if there is no annotated tags and also preserves the `_dirty` suffix. Of course now the `current_commit()` is not really referring to the `current_commit`, so the naming should be discussed also.

What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+* **[DEPRECATED]** `current_commit()` has been deprecated and replaced by
+  `gitdescribe()` which now replaces the output of `git describe` if an
+  annotated tag exists, otherwise it will return the latest commit hash.
+
 # 0.6.0
 * **[BREAKING]** Reworked the way the functions `projectdir` and derivatives work (#47, #64, #66). Now `projectdir(args...)` uses `joinpath` to connect arguments. None of the functions like `projectdir` and derivatives now end in `/` as well, to ensure more stability and motivate users to use `joinpath` or the new functionality of `projectdir(args...)` instead of using string multiplication `*`.
 * New function `parse_savename` that attempts to reverse engineer the result of `savename`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 0.7.0
 * New macro `@savename` that is a shortcut for `savename(@dict vars...)`
-* New function `gitdiscribe` (see below)
+* New function `gitdescribe` (see below)
 * **[DEPRECATED]** `current_commit()` has been deprecated and replaced by
   `gitdescribe()` which now replaces the output of `git describe` if an
   annotated tag exists, otherwise it will return the latest commit hash.
--- a/src/saving_tools.jl
+++ b/src/saving_tools.jl
@@ -12,9 +12,19 @@ with `"_dirty"`.
 
 Return `nothing` if `gitpath` is not a Git repository.
 
-See also [`tag!`](@ref).
+The format of the `git describe` output in general is
+
+    `"TAGNAME-[NUMBER_OF_COMMITS_AHEAD-]gLATEST_COMMIT_HASH[_dirty]"`
+
+If the latest tag is `v1.2.3` and there are 5 additional commits while the
+latest commit hash is 334a0f225d9fba86161ab4c8892d4f023688159c, the output
+will be `v1.2.3-5-g334a0f`. Notice that git will shorten the hash if there
+are no ambiguous commits.
+
 More information about the `git describe` output can be found on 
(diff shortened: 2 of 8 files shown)


tamasgal:
I did not manage to reproduce the errors of the CI on my own machine yet 🤔

Datseris:
Thank you very much for the contribution! I think I like this change and I will review properly once I have some extra free time in a couple of days (currently correcting my thesis :P )

tamasgal:
> Thank you very much for the contribution! I think I like this change and I will review properly once I have some extra free time in a couple of days (currently correcting my thesis :P )

Sure, no hurries ;) I am still writing mine (a few weeks left) 😬

src/saving_tools.jl
```diff
@@ -19,24 +19,35 @@ julia> current_commit()
...
-        @warn "The Git repository is dirty! Adding appropriate comment to "*
-        "commit id..."
-        return c*"_dirty"
+        suffix = "_dirty"
+        @warn "The Git repository is dirty! Appending $(suffix) to the commit ID"
+    end
+    # then we return the output of `git describe` or the latest commit hash
+    # if no annotated tags are available
+    repo = LibGit2.GitRepo(gitpath)
+    c = begin try
```
Datseris: what does the `begin` keyword achieve here?

src/saving_tools.jl
```diff
@@ -19,24 +19,35 @@ julia> current_commit()
...
+julia> current_commit(another_path_with_annotated_git_tags)
+"v1.2.3-g7364ab"
 ```
 """
 function current_commit(gitpath = projectdir())
     # Here we test if the gitpath is a git repository.
     try
         repo = LibGit2.GitRepo(gitpath)
     catch er
-        @warn "The current project directory is not a Git repository, "*
```
Datseris: Here `*` is necessary. If you remove it the string on the line below is not chained to the above and only the first line is displayed as a warning.

src/saving_tools.jl
tamasgal: Ah, that's a leftover, I'll remove it!

src/saving_tools.jl
tamasgal: Yep! That was a typo... thanks!

tamasgal:
> * are the 7-string shortcut strings always equivalent to a commit? From them can you always find the correct commit? (because then I am confused as to why do commit IDs have so much more characters)

As far as I know the abbreviations are guaranteed to be safe, as git will check for possible clashes. Once a clash is detected, the number of characters will increase.
The same holds for all kind of commit hash usages in git (like `git checkout abcd...`) where you can only provide a few characters and git will not complain if it's unique.

> * Indeed the name does not make sense anymore. I propose `gitdescribe` (please deprecate it). But then, in the case where a tag doesn't exist, should we still use the string with 100 characters instead of just 6?

OK `gitdescribe` sounds good to me. I think falling back to the full commit hash is reasonable...

> * `tagsave`, etc. should indeed use `gitdescribe`, provided that it is documented and the way the string works is explained (like e.g. the amount of commits after the tag, etc.)

We can link to the official docs to `git describe`: https://git-scm.com/docs/git-describe and summarise it in a few words I guess?

> * Please add a changelog entry / version update (0.7.0) / docstring

:+1: 

> * Please test the new functions in a case where a commit exists, but no tags. Currently in the tests `current_commit` is tested based on the repo of `DrWatson` that does have a tag. Maybe a quick `initialize_project` and call `gitdescribe` on this project?

For sure :+1:

It will take some time since I am also very busy with my PhD

Datseris:
Alright, I agree with everything you said.

> It will take some time since I am also very busy with my PhD

No problemo, (I just finished the corrections to mine, hooray), just gimme a tag when you think I should merge!

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # 0.7.0
 * New macro `@savename` that is a shortcut for `savename(@dict vars...)`
+* New function `gitdiscribe` (see below)
+* **[DEPRECATED]** `current_commit()` has been deprecated and replaced by
```
tamasgal: `describe` vs. `discribe`

tamasgal:
Ah, thanks for the follow-ups 😃

PR #70, opened 2019-07-18
Add @savename macro

### LINKED ISSUE #60: `@savename` macro
Would be nice to have a macro `@savename a b c` that simply first calls `d = @dict a b c` and then returns `savename(d)`.

sebastianpech:
Implements #60. Turned out to be not just calling `@dict` and then `savename`. The `args` need to be escaped before the `dict` macro call which messes up the variable names.

Closes #60

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 0.7.0
-* New macro `@savename` that it a shortcut for `savename(@dict vars...)`
+* New macro `@savename` that is a shortcut for `savename(@dict vars...)`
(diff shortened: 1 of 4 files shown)


Datseris:
Thank you very much @sebastianpech 

Would you kindly:
* add this to the changelog (under version 0.7)
* add it to the docs (under https://github.com/JuliaDynamics/DrWatson.jl/blob/master/docs/src/name.md#convenience-functions )
* add a crosslink to this in the docstring of `savename` 

And then I will merge!

PR #72, opened 2019-07-23
made changes for new api and improved tests

### LINKED ISSUE #68: update to new API of DataFrames.jl
DataFrames.jl has released a new version.
Functions like `collect_results` throw lots of deprecation warnings now.

We should update this before errors start showing up.

JonasIsensee:
I made the relevant changes. No more deprecation warnings.

Also slightly improved the tests to cover more possibilities.

closes #68

### DIFF
--- a/src/result_collection.jl
+++ b/src/result_collection.jl
@@ -90,7 +90,7 @@ function collect_results!(filename, folder;
     end
 
     n = 0 # new entries added
-    existing_files = get(df, :path, ())
+    existing_files = :path in names(df) ? df[:,:path] : ()
     for file ∈ allfiles
         is_valid_file(file, valid_filetypes) || continue
         #already added?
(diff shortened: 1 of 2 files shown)


Datseris:
@tamasgal  or @asinghvi17  would you like to review this? Otherwise I'll do it.

asinghvi17:
Just looked through the code briefly, but you should probably also review it @Datseris.  One question I had - is there a need for `.=`, or can you just wrap in an array and do `=`?

JonasIsensee:
AFAICT the deprecation warnings of `DataFrames` precisely requested this syntax.
See also
https://discourse.julialang.org/t/release-announcements-for-dataframes-jl/18258/38

Datseris:
Okay I'll try to review but I am clueless about DataFrames so... :D I'll try my best

src/result_collection.jl
```diff
@@ -90,7 +90,7 @@ function collect_results!(filename, folder;
     end
 
     n = 0 # new entries added
-    existing_files = get(df, :path, ())
```
Datseris: I assume that `get(df, :path, ())` was deprecated? Too bad, the alternative `:path in names(df) ? df[:,:path] : ()` is unreasonably verbose :(

JonasIsensee: Sadly it was.

PR #74, opened 2019-08-03
Warn user when projectdir is used and no project is active

### LINKED ISSUE #10: Give warning on `projectdir` if project is standard Julia project
if there is a way to consistently tell if the current project is the "julia default" (whatever that means) then we should warn when the function `projectdir()` is called and showing the default dir.

tamasgal:
This is just a straightforward implementation using the hints in #10.

Closes #10

### DIFF
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -10,7 +10,7 @@ export findproject, quickactivate
 """
     function is_standard_julia_project()
 
-Returns true if a project other than the standard Julia project is active.
+Returns true if the standard Julia project is active.
 """
 function is_standard_julia_project()
     Base.active_project() == Base.load_path_expand("@v#.#")
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -27,7 +27,12 @@ projectdir(args...) = joinpath(projectdir(), args...)
 Join the path of the currently active project with `args`
 (typically other subfolders).
 """
-projectdir() = dirname(Base.active_project())
+function projectdir()
+    if is_standard_julia_project()
+        @warn "Using the standard Julia project."
+    end
+    dirname(Base.active_project())
+end
 projectdir(args...) = joinpath(projectdir(), args...)
 
 
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -8,6 +8,16 @@ export findproject, quickactivate
 @deprecate scriptdir scriptsdir #TODO: remove in next release
 
 """
+    function is_standard_julia_project()
+
+Returns true if a project other than the standard Julia project is active.
+"""
+function is_standard_julia_project()
+    Base.active_project() == Base.load_path_expand("@v#.#")
+end
+
+
+"""
     projectdir()
 Return the directory of the currently active project.
 
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -13,7 +13,7 @@ export findproject, quickactivate
 Returns true if the standard Julia project is active.
 """
 function is_standard_julia_project()
-    Base.active_project() == Base.load_path_expand("@v#.#")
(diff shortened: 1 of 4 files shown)


Datseris:
Cool, thanks! 

Could you please:
1. add some basic tests
2. Increment minor version in `Project.toml`
3. Add this new function to the changelog

src/project_setup.jl
```diff
@@ -7,6 +7,16 @@ export findproject, quickactivate
 
 @deprecate scriptdir scriptsdir #TODO: remove in next release
 
+"""
+    function is_standard_julia_project()
+
+Returns true if a project other than the standard Julia project is active.
```
Datseris: I think it returns `true` if the standarrd Julia project is active, not other than the standard.

tamasgal: Yeah sure! It was a bit late 😉

tamasgal:
Ahm, Travis is calling the test suite with `--project=...`? My dead simple test functions will then of course fail ;)

Datseris:
Notice that you can simply call `Pkg.activate()` and this will activate the default Julia project I think. Maybe it is worth trying? I am not _really_ sure what this will do on travis...

tamasgal:
Yeah let's see, it works locally at least ;)

Datseris:
unfortunately it didnt work :(

tamasgal:
Yeah I noticed but I had no time yet to look at it. It's a pity that it works locally... I'll look at it later, today I really need to write a few pages in my PhD thesis 🙈

asinghvi17:
it's not v#.#, that was just a placeholder :)

tamasgal:
Oh, I was already wondering what kind of string magic that is 😅

asinghvi17:
Well, here's the problem:
> `Base.active_project() = "/tmp/tmpOvxF4V/Project.toml"`

so the test subprocess can't seem to affect the standard project.  I wonder if you could manually activate the standard project using the load_path_expand trick?

Datseris:
Maybe travis doesn't have a "standard project"?

asinghvi17:
Possible but unlikely - try running that in a normal Julia command.

Datseris:
For me it is 
```

julia> Base.active_project()
"C:\\Users\\datseris\\.julia\\environments\\v1.1\\Project.toml"
```

asinghvi17:
I meant in Travis, sorry for the confusion :)

sebastianpech:
I think I read something about that test basically run in a sandboxed mode. Expanding the load path with `@v#.#` should be fine, it's also what you find in the julia `LOAD_PATH` variable.
```
julia> LOAD_PATH
4-element Array{String,1}:
 "@"
 "@v#.#"
 "@stdlib"
 "./modules/"
```

asinghvi17:
Cool, I guess that will have to remain untested.  It _is_ only a warning...

asinghvi17:
We could also test with `withenv`.

sebastianpech:
Isn't `withenv` just used for temporary changes in `ENV`?

sebastianpech:
We could leave the test that shouldn't generate a warning. This way we at least get an error if some of the julia path expansion internals are changed.

Datseris:
Sure, I agree with this as well.

Datseris:
So should we revert the change to `@v#.#` ?

sebastianpech:
Yes. For details you can have a look at [Fredrik's julia con talk](https://youtu.be/q-LV4zoxc-E?list=PLP8iPy9hna6StY9tIJIUN3F_co9A0zh0H&t=490). He explains the expansion of `LOAD_PATH` there.

Datseris:
Urgh, sorry I messed up through GitKraken. Anyways I think current master has all the updates :P

fredrikekre:
The Package itself is the active project on CI.

Datseris:
Thanks for the info!

PR #75, opened 2019-08-05
Fix parse_savename to support unicode characters

sebastianpech:
I've changed string indexing in `parse_savename` to consider that having unicode characters means having invalid string indices.
[https://pkg.julialang.org/docs/julia/THl1k/1.1.1/manual/strings.html#Unicode-and-UTF-8-1](https://pkg.julialang.org/docs/julia/THl1k/1.1.1/manual/strings.html#Unicode-and-UTF-8-1)


sebastianpech:
Will merge with the changes from #74 and update the changelog afterwards.

Datseris:
In general I don't write bugfixes in the changelog, so I don't mind merging this as is.

Maybe one day that drwatson becomes super popular we will have to write them :D

PR #79, opened 2019-08-22
Improve prefix handling and documentation

### LINKED ISSUE #78: savename(datadir(), c) does not work as expected
After the latest changes with #64 and #66 , the following thing happens:
```julia
using DrWatson
c = (T = 100, dt = 0.1, N = 25, mode = "bi")
savename(datadir(), c, "dat")
```
```
"C:\\Users\\datseris\\.julia\\environments\\v1.1\\data_N=25_T=100_dt=0.1_mode=bi.dat"
```
What you will see is that the data directory is not properly chained with the `savename(c)`. This is easily fixed by doing instead `datadir(savename(c))`.

The problem however is that `produce_or_load` does not do this. Instead [as seen in its docs](https://juliadynamics.github.io/DrWatson.jl/dev/save/#DrWatson.produce_or_load), we have:
```
produce_or_load([prefix="",] c, f; kwargs...) -> file, path
```
> Let `s = savename(prefix, c, suffix)`...

---

To resolve this I propose to change `produce_or_load` so that it defines `s` as `joinpath(prefix, savename(c, suffix)`. This also ties nicely with the default prefixes that `savename` supports.

(this is breaking)

Datseris:
Closes #78 and improves `produce_or_load`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 0.8.0
+
 # 0.7.1
 * `projectdir()` now warns if no project (other than the standard one) is
   active
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # 0.8.0
+* **[BREAKING]** : Slightly changed how `produce_or_load` uses `path` and interacts with `savename`, to better incorporate the changes done in version 0.6.0. `prefix` is now also supported.
 
 # 0.7.1
 * `projectdir()` now warns if no project (other than the standard one) is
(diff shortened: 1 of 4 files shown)


sebastianpech:
Looks good. I think I would be a good idea to add `prefix` as kwarg, like `suffix`. Im using `prefix` quite often but not for defining a path.

Datseris:
Hm, you know what? This is actually nice. Yeap, I'll make `prefix ` a kwarg for `produce_or_load`.

sebastianpech:
Nice. One thing that came to my mind today, which would allow us to ditch the path argument, is allowing something like this:

```
datadir("something") do
      # pwd() -> datadir("something")
      produce_or_load(d)
end
```

Datseris:
Hm, I must admit that I don't immediately understand your suggestion. What is this supposed to do? 

I merge this for now and we can discuss your suggestion further.

PR #80, opened 2019-08-23
WIP: Generating and storing a patch

### LINKED ISSUE #36: Include a diff with the saved model run?
When experimenting, I often edit the source-code, to check what changing a parameter does.  
I don't want to ~~check-in~~ commit this code as I'm just trying things out.
(Sure, I should refactor and make that a settable parameters, but life is too short.)  

Also, sometimes I just forget to ~~checkin~~ comit before starting a run.  So, how about storing a diff/patch  with the model results?  Just the output of `git diff` should be fine.

Reproduction would then be:
```
git checkout saved-sha
git apply stored-diff
julia -e ...
```

Edit: not sure where the "check-in" came from.  Probably a flash back to [svn](https://sliksvn.com/support/subversion-basics-using-check-out-update-check-in-commit/), probably because I recently cloned a svn repository...

mauro3:
Fixes #36.

Needs tests and documentation.

### DIFF
--- a/src/saving_tools.jl
+++ b/src/saving_tools.jl
@@ -170,9 +170,9 @@ Dict{Symbol,Any} with 3 entries:
   :x      => 3
 ```
 """
-macro tag!(d, gitpath = projectdir())
+macro tag!(d, gitpath = projectdir(), storepatch = true)
     s = QuoteNode(__source__)
-    :(tag!($(esc(d)), $(esc(gitpath)), $s))
+    :(tag!($(esc(d)), $(esc(gitpath)), $(esc(storepatch)), $s))
 end
 
 """
--- a/src/saving_tools.jl
+++ b/src/saving_tools.jl
@@ -91,9 +91,9 @@ end
 
 """
     tag!(d::Dict, gitpath = projectdir()) -> d
-Tag `d` by adding an extra field `commit` which will have as value
+Tag `d` by adding an extra field `gitcommit` which will have as value
 the [`gitdescribe`](@ref) of the repository at `gitpath` (by default
-the project's gitpath). Do nothing if a key `commit` already exists or
+the project's gitpath). Do nothing if a key `gitcommit` already exists or
 if the Git repository is not found.
 
 Notice that if `String` is not a subtype of the value type of `d` then
@@ -110,7 +110,7 @@ Dict{Symbol,Int64} with 2 entries:
 julia> tag!(d)
 Dict{Symbol,Any} with 3 entries:
   :y      => 4
-  :commit => "96df587e45b29e7a46348a3d780db1f85f41de04"
+  :gitcommit => "96df587e45b29e7a46348a3d780db1f85f41de04"
   :x      => 3
 ```
 """
@@ -119,19 +119,19 @@ function tag!(d::Dict{K, T}, gitpath = projectdir(), source = nothing, storepatc
     c = gitdescribe(gitpath)
     patch = gitpatch(gitpath)
     c === nothing && return d # gitpath is not a git repo
-    if haskey(d, K("commit"))
-        @warn "The dictionary already has a key named `commit`. We won't "*
+    if haskey(d, K("gitcommit"))
+        @warn "The dictionary already has a key named `gitcommit`. We won't "*
         "add any Git information."
         return d
     end
     if String <: T
-        d[K("commit")] = c
+        d[K("gitcommit")] = c
         if patch!=""
             d[K("gitpatch")] = patch
         end
     else
         d = Dict{K, promote_type(T, String)}(d)
-        d[K("commit")] = c
+        d[K("gitcommit")] = c
         if patch!=""
             d[K("gitpatch")] = patch
         end
@@ -164,7 +164,7 @@ julia> d = Dict(:x => 3)Dict{Symbol,Int64} with 1 entry:
 
 julia> @tag!(d) # running from a script or inline evaluation of Juno
 Dict{Symbol,Any} with 3 entries:
-  :commit => "618b72bc0936404ab6a4dd8d15385868b8299d68"
+  :gitcommit => "618b72bc0936404ab6a4dd8d15385868b8299d68"
   :script => "test\\stools_tests.jl#10"
   :x      => 3
 ```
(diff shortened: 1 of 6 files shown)


src/saving_tools.jl
```diff
@@ -103,9 +118,15 @@ function tag!(d::Dict{K, T}, gitpath = projectdir(), source = nothing) where {K,
...
     if String <: T
         d[K("commit")] = c
+        if patch!=""
+            d[K("gitpatch")] = patch
+        end
     else
         d = Dict{K, promote_type(T, String)}(d)
         d[K("commit")] = c
+        if patch!=""
+            d[K("gitpatch")] = patch
```
Datseris: Why are you naming this `gitpatch` instead of `gitdiff`? It does directly contain the output of `gitdiff`, correct?

README.md
```diff
@@ -1,3 +1,5 @@
+test
```
Datseris: ? :D

src/saving_tools.jl
```diff
@@ -65,6 +65,20 @@ function gitdescribe(gitpath = projectdir())
...
+
+Generates a patch describing the changes of a dirty repository
+compared to its last commit.
+"""
+function gitpatch(gitpath = projectdir())
+    # tree = LibGit2.GitTree(repo, "HEAD^{tree}")
+    # diff = LibGit2.diff_tree(repo, tree)
+    # now there is no way to generate the patch with LibGit2.jl.
+    # Instead use commands:
+    patch = read(`git --git-dir=$(gitpath)/.git diff `, String)
```
sebastianpech: Maybe a `git diff HEAD` would be better here to also include staged files

sebastianpech: I mean, it depends on what you want actually. If you want to diff the index, than you version is fine. However, the docstrings says diff the last commit, so you need `HEAD`.

README.md
mauro3: Oops... I needed to make a dirty repo, so that's what I did ;-)

src/saving_tools.jl
mauro3: Well, the diff(erencing) produces a patch.  But I'm easy.

src/saving_tools.jl
mauro3: You're right, thanks!

src/saving_tools.jl
Datseris: Alright, let's keep it patch then. I'm also easy on this.,

Datseris:
Alright, the output is human readable once you call `print` on it, so that `\n` and stuff are rendered properly. Once this has tests/docs we can merge! :)

src/saving_tools.jl
mauro3: There is the other inconsistency: "gitpatch" vs "commit" but I felt that "patch" alone was a bit confusing.  Maybe change "commit" -> "gitcommit"?

Datseris: hm, good point. Although breaking I think this is a good improvement (and the next release has one other breaking change anyways so let's go with it!)

mauro3:
Ok, I'm mystified by two test failures:
- one on my local machine, identical to https://github.com/JuliaIO/JLD.jl/issues/205.  Has anyone experienced this before?
- and the OSX one on travis: https://travis-ci.org/JuliaDynamics/DrWatson.jl/builds/575994409?utm_source=github_status&utm_medium=notification

Any pointers welcome.

ToDo: docs.  I'll just add them to the `tag!` docstring. Anywhere else?

Side note: if a binary file is different, this will not work and will actually make the whole patch not applicable (although the patch can be hand edited to make it applicable to non-binary files).  We could add support for binary files by running `git diff --binary HEAD` but that has the potential to produce very large diffs.  We should probably cross this bridge when we get there.  Until then I'll add a note to the docs.

mauro3:
Ok, I did the rename "commit" -> "gitcommit".  Let's also see whether that OSX failure was fluke.

Datseris:
> Side note: if a binary file is different, this will not work and will actually make the whole patch not applicable (although the patch can be hand edited to make it applicable to non-binary files). 

This is not clear for me. Let's say I have a dirty repo with a change in a text file and a binary file. What happens at the moment out of the following?

1. No patch is produced (`nothing`).
2. A patch is produced, for the text files only.

(let's **not** make patches for binary files, which are non-human-readable)

This brings an interesting question though... what is the typical size of the saved `String` from this patch? Most of the time it will be 10s pages of text. How much larger does this make the file I wonder?

mauro3:
For a change in a binary and a text file, the patch looks like so:
```
diff --git a/shot0001.png b/shot0001.png
index 5fc9d26..8e4091e 100644
Binary files a/shot0001.png and b/shot0001.png differ
diff --git a/text b/text
index 94cb47e..7d8fcab 100644
--- a/text
+++ b/text
@@ -1,4 +1,6 @@
 adsflk
 dsflkj
 
+add
+
 adsflkj
\ No newline at end of file
```
However, apply this does not work as it cannot apply the binary part.  Editing it (by hand) to
```
diff --git a/text b/text
index 94cb47e..7d8fcab 100644
--- a/text
+++ b/text
@@ -1,4 +1,6 @@
 adsflk
 dsflkj
 
+add
+
 adsflkj
\ No newline at end of file
```
does then work.

Datseris:
Ah I see! Okay, this is fine for me, I'd say to leave the situation as is. Just put a note as you suggested. For docs, I think just `tag!` is fine! I'll re-arrange the docs a bit after this pr.

mauro3:
Added docs to `tag!`.  If the test failure on OSX is ok, then this is ready to merge after a look-over the docs.  If that failure is not ok, then I need some help as I don't have the time to go bug hunting on travis right now.

Datseris:
Ah tests fail because of:
```
 Expression: "script" ∈ keys(file)

   Evaluated: "script" ∈ ["gitcommit", "b", "simulation", "gitpatch", "a"]
```
Did you change the functionality of `@tag` to not include the script maybe?

mauro3:
Yes, I re-ordered the arguments of `tag!` to keep `source` last.  Combined with my unrelated local testfailures, I only figured this out now.  Maybe better luck now...

Datseris:
Perfect, I'll merge tomorrow morning (just arrived from a 12 hour flight, my brain hurts); I'll check the docs and do any improvements if need be, but I don't see any.

mauro3:
Aright, I'll see what I can do tonight, but my brain is a bit tired too... You have a good rest!

PR #81, opened 2019-08-25
Replace GitRepo -> GitRepoExt to get repo also when using sub-dirs

mauro3:
When doing, e.g., `gitdescribe(".")` and "." is a subdirectory to the topmost gid-dir (where `.git` is), then currently:
```
julia> gitdescribe(".")
┌ Warning: The directory ('.') is not a Git repository, returning `nothing` instead of the commit ID.
└ @ DrWatson ~/.julia/dev/DrWatson/src/saving_tools.jl:46
```
With this PR it "correctly" returns the hash of the repo.

As I haven't gotten around to trying DrWatson out, I don't know whether this is an improvement or not.  Also, if #80 gets merged, one more `GetRepo` -> `GetRepoExt` needs to be done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# 1.0.0
+* **[BREAKING]** : The `gitpath` argument used among many functions
+  can now also point to a subdirectory within a git repository.
+  Previously it had to be the top directory (i.e. the one containing
+  `.git/`).
+* For a dirty repository, `tag!` and friends now store a patch (aka a
+  diff) of the changes.  Thus even not-commited code changes can now
+  reproduced.  Implemented in the `gitpatch` function.
+
 # 0.8.0
 * **[BREAKING]** : Slightly changed how `produce_or_load` uses `path` and interacts with `savename`, to better incorporate the changes done in version 0.6.0. `prefix` is now also supported.
 * `tag!` and co now also store the git diff patch if the repo is dirty (#80).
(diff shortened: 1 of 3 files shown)


mauro3:
On my local machine tests don't pass because I hit something like JuliaIO/JLD.jl#205, so it's a bit hard to figure out where to adjust them... but if this PR is welcome, I'll do the leg-work.

Datseris:
tests pass here! I think this change is fine, as (I assume) works properly when you have a git directory inside another git directory. If you call `gitdescribe(".")` with `.` the subdir, then even with the current change it will properly return the subdir git repo, correct?

Datseris:
(ah but maybe we should document this change, it may be unexpected behavior if not documented)

mauro3:
Ok, let me know where to document it. In the CHANGELOG presumably.  Also in the doc-string?

Datseris:
Okay #80 is in, you can update this now. I'd document this in the docstring of both `gitdescribe` and `gitpatch`. An entry in the changelog would be appreciated as well! :)

mauro3:
Somehow the Travis link disappeared, probably because I pushed the CHANGLOG and doc-changes with `[ci skip]`.  The build is https://travis-ci.org/JuliaDynamics/DrWatson.jl/builds/578947414.

I also added #80 to the changelog.

Datseris:
0.8.0 is not actually tagged yet, I was waiting for this, so I merged the changelogs :) 

Thanks a lot for this!

Issue #67, opened 2019-07-15
Running project tests


sebastianpech:
I'd like to discuss how to nicely run tests defined for a DrWatson project.

- So by default we have the test directory containing a `runtests.jl`. This matches the default for julia packages / projects. 
- `Pkg.test` can be invoked without additional arguments, which results in running the tests for the currently active project.

So in a freshly initialized DrWatson project
```
julia> using DrWatson

julia> quickactivate(@__DIR__)

(watson test) pkg> test
ERROR: The following package names could not be resolved:
 * watson test (not found in project or manifest)
Please specify by known `name=uuid`.
```
After adding a UUID to `project.toml`
```
(watson test) pkg> test
   Testing watson test
ERROR: expected the file `src/watson test.jl` to exist for package watson test at /Users/sebastianpech/test/julia/watson test
```
After adding the file `src/watson test.jl`:
```
(watson test) pkg> test
   Testing watson test
 Resolving package versions...
   Testing watson test tests passed
```
The tests run successfully.

What do you think about that approach? Alternatively we could add a function like `quickruntests(@__DIR__)` or something similar which allows us the keep the current project structure.

asinghvi17:
I think we could probably initialize it with a src/PkgName.jl and a UUID, wouldn't be too hard a change.

Datseris:
Damn. :( The problem you mention here is the same as here: https://github.com/JuliaDynamics/DrWatson.jl/issues/61

---

Here is why I don't like the UUID approach:
From my perspective a project created by DrWatson is not supposed to be a Julia package, which is why it doesn't have a UUID. It is supposed to be a Julia project. I believed these two concepts are clearly distinguished, but if you don't think so maybe I should reconsider...

I also don't think creating a `PackageName.jl` -like file makes sense. A scientific project more often that not will try to implement several functionalities and pipelines to answer the scientific questions. I doubt this would typically be possible to do with a single module named `PackageName`. See the discussion here as well : https://github.com/JuliaDynamics/DrWatson.jl/issues/62

Such discussions like the one here is also why I was very skeptical of: https://github.com/JuliaDynamics/DrWatson.jl/issues/40 (although I never expressed the skepticism back then) , because it also tries to make the project initialized by DrWatson more like a Julia package.

---

[...]

PR #90, opened 2019-09-25
Remove test directory from `initialize_project`

Datseris:
Well, it just so happens that every month I find a new reason to regret merging #43. No more I will live with such regrets, I am reverting it!

---

The motivation is simple: this "test" directory is clearly targeting a Julia package. I am not sure how many scientists write consistent checks for their actual projects. I am for sure not one of them. The advantage of not having a test directory will be great: hopefully people will not tend to think that DrWatson is about Julia packages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.1.0
+* `initialize_project` no longer makes a test directory.
+
 # 1.0.1
 * Allow `tag!` and derivatives to handle dictionaries with *key type* `Any`.
 # 1.0.0
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.0.1"
+version = "1.1.0"
 
 
 [deps]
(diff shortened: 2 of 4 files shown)


asinghvi17:
I have noticed that it's a bit hard to write tests of experimental code 😛

jessebett:
Hey @Datseris, 

Sad to have seen this get unmerged. Writing unit tests for scientific code is good for making sure aspects of your inference pipeline haven't changed in ways that break your previous assumptions.

tamasgal:
We had long discussions about this and yes, DrWatson creates a project folder and not a Julia package, so we decided to not confuse the package testing mechanism with it.

Another point is that usually one creates Packages to consolidate functionalities which are then documented and tested. A DrWatson protect is therefore more like a collection of composition of such packages so unit tests in that sense are out of the scope. We are more likely talking about high or even higher level tests and regression tests which needs to be individually tailored. 

That being said, nothing prevents you from adding your own test folders in whatever structures you like and you can add test procedures for all kinds of cases but we simply could not come up with a common structure which fits most cases and does not confuse people.

Datseris:
Sorry for the dissatisfaction Jesse, but sometimes decisions have to be made, and this made DrWatson appear too much like a package generator or package development tool, which is not. I made the choice in favor of reducing confusion with the result of increasing dissatisfaction in a (hopefully small) percentage of users. You can't always win...

I just want to point out a technical detail: https://github.com/JuliaDynamics/DrWatson.jl/issues/67 

Simply put, you cannot use the command `pkg> test YourScientificProject`.

PR #92, opened 2019-10-01
Improve `savename` for nested containers (default_expand)

Datseris:
This PR significantly improves the behavior of `savename` with nested containers that are expanded with `default_expand` or other means. I found out that this is actually something I use a lot in my work, but it was by far not so nice. 

The main problems so far was that nested containers (that were supposed to be printed) were printed if they were empty, which led to me having stuff like: `Barkley_T=100_p=().bson`. Here having the `p` is meaningless, since its default value is an empty named tuple. If instead I had `p = (a=5,)` then I got the nice `Barkley_T=100_p=(a=5).bson`, as expected.

The second problem was that the parameter container `p` that is expanded, could be expanded to an empty `savename`, because it contained e.g. a matrix. This is also solved.

(see the tests for confirmation)


Datseris:
```julia
struct A
    a
    p
end
DrWatson.default_allowed(::A) = (Any,)
DrWatson.default_expand(::A) = ["p"]

x = A(5, (b = 3, c = 4))
savename(x) == "a=5_p=(b=3,c=4)"

x = A(5, NamedTuple())
savename(x) == "a=5"

x = A(5, (b = rand(50,50),))
savename(x) == "a=5"
```

Datseris:
Although of low priority in my eyes, it could be straight-forward. Nested containers are always put inside parenthesis `()` with a raw `savename` call. So after the identification of the `=` and the `_` another step would identify possible `()` and call `parse_savename` directly on this.

Issue #84, opened 2019-09-02
Keyword arguments for tag! and tagsave


Datseris:
The arguments of `tag!` and `tagsave` have become quite extensive. Maybe it is time to define keyword-based versions of these functions. This is very easy and also non-breaking.

Here is where is `tag!` defined: https://github.com/JuliaDynamics/DrWatson.jl/blob/master/src/saving_tools.jl#L125

I don't really know how keywords work with macros, so I don't know how `@tagsave` can be affected.

mauro3:
> I don't really know how keywords work with macros, so I don't know how @tagsave can be affected.

You got to be a bit careful because `:(a = 1)` parses as an `Expr(:=,:a,1)` whereas for kws you need a `Expr(:kw, :a, 1)`. So some extra processing is probably needed but nothing too tricky.

sebastianpech:
@Datseris you mean additional kw-methods for `tag! `and `tagsave`? So one should still be able to use the non-kw version?

Datseris:
yeap, because this is the only way to make it non-breaking (since we are now in 1.0! :) )

sebastianpech:
Thought so. Kinda tricky to not get warnings, but I figured it out. Will make a PR tomorrow.

Issue #83, opened 2019-09-02
[FR] "force" argument for tag! and derivatives


Datseris:
At the moment `tag!` (and thus all derivatives up to `@tagsave`) do not do anything if the given dictionary already has fields like `gitcommit` etc. This is done for safety purposes.

However, I now see that an argument `force` would be helpful, so that a dictionary that is already tagged can be "retagged". This is useful in the following scenario: when there is a processing pipeline that adds data to an existing dictioanry, which is saved on disk. The dictionary is updated as new experimental data are accessible. I am now in exactly this scenario and I can see how this is useful.

As this can only be a non-breaking change, since we are now in 1.0, I'd suggest to do this after #84 , and simply make `force` a keyword argument.

Datseris:
I am tagging this as "low-priority" because one can always manually call `delete!` on the dictionary (which is what I do now).

PR #93, opened 2019-10-03
Improve tag functions and macros

sebastianpech:
Implements #84. I'll also add #83, so this is still a draft.

The problems with having kw and non-kw versions are the following:

**Methods:** The definition of the kw-function also creates a method without the keywords. However, this method is also defined by the optional args function defintion. So on `using` you get a warning that the method is overwritten. My solution here is to define one additional argument, in the non-kw function, as mandatory. So dispatch works like this now:
- If this argument is not given -> dispatch on the kw-version
- If it is given -> dispatch on the non-kw-version

**Macros:** This one is a little tricky. During dispatch one can't tell if the passed `args...` to the macro are kw-definitions or not eg:
```
julia> :(@__DIR__)|>typeof
Expr
julia> :(x=10)|>typeof
Expr
```
The solution now basically works like this.
1. Collect mandatory and optional arguments
2. Check if all optional arguments could be kw-definitions
3. If yes, splat them as keywords
3. If no, splat them in order they were given and then splat the remaining optional arguments, if any.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.3.0
+* The functions `tag!, tagsave` and their respective macros now obtain their arguments (besides the first two) as keywords instead of positional arguments. The positional versions are deprecated (#93).
+* New keyword `force = false` for `tag!` and co. which replaces the existing `gitcommit` field.
+
 # 1.2.0
 * Improved behavior of `savename` with respect to nested containers. If a nested container is empty, it is not printed instead. For example, `T=100_p=()_x=2` now becomes `T=100_x=2`. (if `p` is not empty then it is expanded as usual)
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.2.0"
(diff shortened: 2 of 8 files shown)


sebastianpech:
Any clue why it sometimes failes on mac? It looks pretty random to me, though it seems to always fail with this error:
```
Project Setup: Test Failed at /Users/travis/build/JuliaDynamics/DrWatson.jl/test/project_tests.jl:45
  Expression: projectname() == name
   Evaluated: "test project" == "lala"
```
Is there some async action going on in `initialize_project` such that sometimes the folder is not yet created?

Datseris:
Yeah, the macro version is tough, but you don't have to do this in this PR. If you feel it is difficult or too tricky, just drop it and we can have a look later.

> My solution here is to define one additional argument, in the non-kw function, as mandatory. 

I agree. Maybe it is a good opportunity to *deprecate* the non-kw version?

I don't know about the error you cite. I have seen it before and the solution for was always to just restart tests :P Maybe its time to be opened as a formal issue? but still, I don't really know why it happens :P

sebastianpech:
Ok. Strange. Macro version is done ;)

Datseris:
It would be tricky adding a deprecation warning here, since the low-level method is the non-kw. Instead let's just replace the documentation string to display only the keyword version.

sebastianpech:
Oh yes. But I can just change the call order

sebastianpech:
Thanks for updating the changelog

PR #96, opened 2019-10-04
Tests failing "randomly" on mac CI

sebastianpech:


sebastianpech:
I think I found the problem. The old method of obtaining the projectname through `promptf` involved a check of `mtime` in [Pkg](https://github.com/JuliaLang/Pkg.jl/blob/a10291a94d2e956dad980ccdc4dba17b1cf274e5/src/REPLMode/REPLMode.jl#L488)
```
[...]
 if prev_project_file == project_file && prev_project_timestamp == mtime(project_file)
                prefix = prev_prefix
[...]
```
Due to the use of force, the first condition was always true and somehow (don't know if coincidentally) on osx the timestamp didn't change. So `promptf` returned the old prefix.

tamasgal:
Ah wow, I also had some trouble in past with `mtime` on macOS 🙈

sebastianpech:
However, I'm puzzled how this can happen. Is it possible, that from initializing the first project to initializing the second one with force (and inbetween deleting all files in the projectdir) less than 1/100 of a second passed?

sebastianpech:
Nope ...
```
julia> function testmtime()
           touch("testfile")
           t1= mtime("testfile")
           rm("testfile")
           touch("testfile")
           t1, mtime("testfile")
       end
testmtime (generic function with 1 method)

julia> testmtime()
(1.570177477413133e9, 1.570177477413308e9)
```

tamasgal:
The problems I encountered in past were more low level, involving `atimes` and `mtimes` on the file system level. Not sure if this is related but it was similarly unpredictable...

sebastianpech:
Ok. Yes seems to be on a lower level. On my device I never had problems with that.

Datseris:
wow amazing, thanks !!!! You're really pro to figure out this!

sebastianpech:
Thanks ;)

Issue #63, opened 2019-07-04
savename digits respect "1e-6" syntax


jessebett:
If I specify an argument like `:lr=>1e-5` I would like `savename` to not attempt to round this. At the moment this is beyond the default `digits=4` and gets saved as `lr=0`. 

Is it easy to distinguish floats written as `1e-2` from `0.01`? If so, can the former be preserved in the savename?

sebastianpech:
I don't think that this is possible. `1e-5` is just a different method of defining a float, so the calling function justs sees the float and not the used notation. 

Alternatively we could check the rounding results in `savename` and if it's `0.0` instead display `digits` significant digits.

Julia internally uses the [grisu algortihm](https://github.com/JuliaLang/julia/blob/55e36cc308b66d3472990a06b2797f9f9154ea0a/base/grisu/grisu.jl) for printing floating point numbers. 
There is `Base.Grisu.print_shortest` which returns the shortest possible representation of a floating point number while alllowig parsing the resulting string to the same number.

Datseris:
Alright, I don't see this getting into 1.0, but this can only be accepted as a non breaking change. Here is my proposal: 

Implement a keyword argument `scientific = nothing`. If `nothing`, the behavior is as is. If it is an integer, the result is plotted in scientific notation, with `scientific` significant digits being displayed. E.g. `0.00123` becomes `123e-5` if `scientific` is 3 or more, but if e.g. `scientific` is 1, then it becomes `1e-3`.

PR #97, opened 2019-10-08
Add scientific keyword to savename

sebastianpech:
Implements #63

### DIFF
--- a/test/naming_tests.jl
+++ b/test/naming_tests.jl
@@ -88,12 +88,12 @@ c = 1
 d = "test"
 di = @dict a b c d
 
-@test savename(di,scientific=6) == "a=1.2345e-7_b=1.0_c=1_d=test"
-@test savename(di,scientific=5) == "a=1.2345e-7_b=1.0_c=1_d=test"
-@test savename(di,scientific=4) == "a=1.234e-7_b=1.0_c=1_d=test"
-@test savename(di,scientific=3) == "a=1.23e-7_b=1.0_c=1_d=test"
-@test savename(di,scientific=2) == "a=1.2e-7_b=1.0_c=1_d=test"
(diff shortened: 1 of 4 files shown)


test/naming_tests.jl
```diff
@@ -80,3 +80,22 @@ x = A(5, NamedTuple())
...
+c = 1
+d = "test"
+di = @dict a b c d
+
+@test savename(di,scientific=6) == "a=1.2345e-7_b=1.0_c=1_d=test"
+@test savename(di,scientific=5) == "a=1.2345e-7_b=1.0_c=1_d=test"
+@test savename(di,scientific=4) == "a=1.234e-7_b=1.0_c=1_d=test"
+@test savename(di,scientific=3) == "a=1.23e-7_b=1.0_c=1_d=test"
+@test savename(di,scientific=2) == "a=1.2e-7_b=1.0_c=1_d=test"
+@test savename(di,scientific=1) == "a=1.0e-7_b=1.0_c=1_d=test"
```
Datseris: In the case of `scientific = 1`, isn't the second digit always `0` and thus can be skipped?

sebastianpech: Yes it is. Do you think it's better to fallback to the rounding behavior of the default rounding method? So also adding the check with Integer rounding.

sebastianpech: In favor of consistency it's better to remove it

Datseris: > Yes it is. Do you think it's better to fallback to the rounding behavior of the default rounding method?

Yes I think so.

Datseris: (to be clear I think that `1.0e-7` should be written as `1e-7`. But if this is too hard to do programatically I can live with it not happening.

sebastianpech: Ah, now I get it. That shouldn't be too difficult.

PR #100, opened 2019-10-11
Only use major.minor.patch for compat version

### LINKED ISSUE #99: compat in Project.toml breaks on rc versions
PR #94 includes a line with the Julia version for compat info. However, if the Julia version is anything other that x.y.z (for example, the rc version `1.3.0-rc2.0` I'm currently using) this breaks because `Pkg` can't/won't parse that string.

I'm not sure if this should be fixed in Pkg (I'm assuming it won't because I don't think rc versions are included into SemVer?) or in DrWatson. I can try to provide a PR if the decision is made for the latter.

karajan9:
Fixes #99

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.4.1
+* Fix a bug that created incompatible version strings in generated `Project.toml` files on release candidate versions of Julia.
+
 # 1.4.0
 * `savename` now supports rounding to significant digits with the keyword argument `scientific`, where `scientific` defines the number of significant digits.
 # 1.3.0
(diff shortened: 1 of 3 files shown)


Datseris:
Yes I am fine with this. I consider #99 a bug in any case. Please increment our `Project.toml` version by a bugfix number, and if you are willing you may add a comment in Changelog.md!

karajan9:
Done!

karajan9:
Maybe it would make sense to also do CI on the DEV version of Julia? Besides (maybe?) catching this bug, future problems can be anticipated earlier.

Issue #105, opened 2019-11-05
Overload `save` and `load` instead of aliasing?


torfjelde:
Seems to me that `wsave` and `wload` is used everywhere internally for saving and loading files. At the moment these are just aliases for  `FileIO.save` and `FileIO.load` respectively:

https://github.com/JuliaDynamics/DrWatson.jl/blob/28ac0a17c846ab2d96e08b713755ba22805f904a/src/DrWatson.jl#L20-L21

Would it be a good idea to instead do something like
```julia
wsave(args...) = save(args...)
wload(args...) = load(args...)
```
so that the user can overload the saving behavior used internally without overloading `FileIO.save` and `FileIO.load`? E.g. might want to save plots using `Plots.save`, in which case could just implement `wsave(p::Plots.Plot, f) = Plots.savefig(p, f)` and thus, for example, get `safesave` for "free" rather than having to overload `safesave` itself.

I'll can make a PR if this sounds like a good idea.

Datseris:
Yeap, this sounds a good idea, and should have been the original intent.

The reason I wrote `wsave` and `wload` is so that in the initial stages of development we were not bound to FileIO. But you seem to have an excellent point as well :) 

If you notice, other functions like `tmpsave` and `tagsave` etc all call `mkpath` internally and then finally call `wsave`. A good progress would be to make `wsave` call `mkpath` and then call `FileIO.save` (and also overload things like saving plots).

If you're up for making this change that would be awesome.

Issue #85, opened 2019-09-03
Integration with ploting


mauro3:
It would be nice if figures produced from simulations had the info to reproduce them associated with them.  Maybe as metadata tags (which most formats support)?  I'm just putting this out here as today I inherited a bunch of figures which I wouldn't know where they came from.

asinghvi17:
Perhaps just using savename to save them is also possible?

Datseris:
This has been discussed a bit in the announcement post. See here : https://discourse.julialang.org/t/drwatson-the-perfect-sidekick-to-your-scientific-inquiries/23241/6?u=datseris and below

sebastianpech:
I use savename for this. However, finding a general solution to this would actually be cool. What is the current status of the chaosdb implementation?

Datseris:
> What is the current status of the chaosdb implementation?

Not clear at the moment... The authors haven't worked on it for some months now, since they got a little baby :) I have urged them to set up some "basic usage and installation" instructions so that I can help them progress, as I don't know how to use CaosDB on windows at the moment, but I am still waiting on this. @salexan2001 any updates?

But I am not so sure whether this will solve the current issue, which (i believe) is mostly requesting the `tag!` functionality but for figure objects. I think the ideas in the discourse post may be more fitting. To be more exact, CaosDB *can* solve this issue, but it may be more involved than what mauro is asking for.

sebastianpech:
Ok, thanks for the update.

> But I am not so sure whether this will solve the current issue, which (i believe) is mostly requesting the tag! functionality but for figure objects. I think the ideas in the discourse post may be more fitting.

Yes, you're right. It's just a topic that keeps bothering me.

salexan2001:
After my parental leave I have restarted working on the Julia interface for CaosDB. We were encountering many problems with several XML libraries for Julia, so we are working on JSON support of the protocol during the next weeks. Apart from that, the interface is not relying on the C library anymore which makes usage on other platforms easier.

Datseris:
I was sick and tired of constantly having PyPlot complaining that the path that I want to save my figure doesn't exist, I took advantage of the advances we had on defining `wsave` for custom types. I now do
```julia
DrWatson._wsave(s, fig::Figure) = fig.savefig(s, dpi = 300, transparent = false)
```
which gives me the functionality of making the path first, as well as using `safesave` for free!

PR #108, opened 2019-11-05
`wsave` and `wload` can now be overloaded

torfjelde:
Addresses #105

### DIFF
--- a/src/DrWatson.jl
+++ b/src/DrWatson.jl
@@ -17,11 +17,10 @@ include("saving_tools.jl")
 # Functionality that saves/loads
 using FileIO: save, load
 export save, load
+
+_wsave(filename, obj) = save(filename, obj)
 wsave(args...) = save(args...)
-function wsave(filename, obj)
-    mkpath(dirname(filename))
-    save(filename, obj)
-end
+wsave(filename, obj) = (mkpath(dirname(filename)); _wsave(filename, obj))
 
 wload(args...) = load(args...)
 
--- a/src/DrWatson.jl
+++ b/src/DrWatson.jl
@@ -15,7 +15,7 @@ include("naming.jl")
 include("saving_tools.jl")
 
 # Functionality that saves/loads
-using FileIO: save, load
+using FileIO
 export save, load
 export wsave, wload
 
--- a/src/DrWatson.jl
+++ b/src/DrWatson.jl
@@ -17,12 +17,18 @@ include("saving_tools.jl")
 # Functionality that saves/loads
 using FileIO: save, load
 export save, load
+export wsave, wload
 
-_wsave(filename, obj) = save(filename, obj)
-wsave(args...) = save(args...)
+_wsave(filename, obj) = FileIO.save(filename, obj)
+
+"""
+    wsave(filename, obj)
+Save `obj` at `filename` via `FileIO` by first creating the appropriate paths.
+"""
 wsave(filename, obj) = (mkpath(dirname(filename)); _wsave(filename, obj))
 
-wload(args...) = load(args...)
+"Currently equivalent with `FileIO.load`."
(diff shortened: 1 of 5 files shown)


src/DrWatson.jl
```diff
@@ -17,8 +17,13 @@ include("saving_tools.jl")
 # Functionality that saves/loads
 using FileIO: save, load
 export save, load
-wsave = save
-wload = load
+wsave(args...) = save(args...)
+function wsave(filename, obj)
+    mkpath(dirname(filename))
+    save(filename, obj)
+end
```
Datseris: Hm, maybe we can do this even better, so that `wsave` always calls `mkpath` regardless of arguments. At the moment if someone overloaded `wsave(::String, ::Plots.Plot, ...)` then they wouldn't get the automatic `mkpath` with it.

Hm, but this seems tricky... I've been thinking about it for two minutes but can't figure a simple way!

Datseris:
This is great. I am assigning myself as well, to add some documentation commits highlighting that by overloading `wsave` one gets `tagsave, safesave` for free.

src/DrWatson.jl
torfjelde: Good point; could of course do something like
```julia
wsave(filename, obj) = (mkpath(dirname(filename)); _wsave(filename, obj))
```
and then have the user override `_wsave`. Not as nice-looking though :/

Datseris: Ah, I see your point. I didn't consider the scenario where someone would be using `wsave` directly, but rather only the higher level functions `safesave` or `tagsave`.

What is the use-case of using `wsave` directly, instead of e.g. `save`?

I am now also thinking about how would `tagsave` work with overloading `wsave` to save figures... How do you add git id to a figure?

torfjelde: I think for the same reasons why we'd use it internally; it's more convenient to write `wsave` than `mkpath(...); save(filename, obj)`:) Also, maybe in the future more functionality is added to `wsave`?

More importantly, if the user wants to extend behavior for other functions using `wsave`, e.g. `tagsave`, the user ends up working with `wsave` instead of `_wsave`. 

But of course, this is all a matter of opinion. I just know that I like the approach taken by Distributions.jl with overloading `_logpdf` once and then you use `logpdf` everywhere else, knowing that certain checks are performed before passing to `_logpdf`. `wsave` is likely to see less frequent use compared to `logpdf` though:)

torfjelde: > I am now also thinking about how would tagsave work with overloading wsave to save figures... How do you add git id to a figure?

This is basically the reason why I started looking into the internals of DrWatson:) I also briefly spoke to @JonasIsensee about this on Slack. I've looked at creating a project similar to DrWatson myself some time back, and I was planning on going for an approach where I have a simple `filename → metadata` lookup-table (started out as a simple json, but played with the  thought of eventually turning it into a SQLite table). You can then just add whatever information you want to any file, e.g. commit id.

torfjelde: Other use-cases for this is when you have maybe have files too large to fit into memory, and so you want to read large files during whatever computation you're doing. Then it would be beneficial to have files in a parquet-format. You might still want to know which commit id produced it

Datseris: Okay, I see your point(s).

Okay then, go ahead and make the change with `_wsave` and we can merge this!

Datseris: related: https://github.com/JuliaDynamics/DrWatson.jl/issues/85

torfjelde: Great! I'll do that asap:)

Datseris:
@torfjelde I've added some docs, have a read and let me know if this makes sense.

torfjelde:
> @torfjelde I've added some docs, have a read and let me know if this makes sense.

Looks great to me!:)

PR #110, opened 2019-11-15
remove infinite spam from QUACKTIVATE and add @quickactivate

### LINKED ISSUE #109: Make `quickactivate` do nothing if the project is already active.
I find it ridicoulous to be constantly spammed with
```
Activating environment at `C:\Users\datseris\ownCloud\MyScience\Projects\Klein Tunneling in Graphene\kwant_graphene\Project.toml`
```
when this project is already active...

I personally find `quickactivate` super convenient, so every script I ever write starts with this function.

Maybe it is a good idea to make it do nothing if the project-to-be-activated is already active? Or maybe there is some drawback in this?

So what I am asking is whether there is any benefit whatsoever into activating again the already active project.

Datseris:
Closes #109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 1.5.1
+* `quickactivate` doesn't do anything anymore if you try to activate to already active project.
 # 1.5.0
 * Started to add support for overloading save/load for custom files. See the updated docs around `wsave` and `wload`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
-# 1.5.1
+# 1.6.0
 * `quickactivate` doesn't do anything anymore if you try to activate to already active project.
+* New macro `@quickactivate`
+
 # 1.5.0
 * Started to add support for overloading save/load for custom files. See the updated docs around `wsave` and `wload`.
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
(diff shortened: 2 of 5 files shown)


Datseris:
I think when I'll give a talk about DrWatson I will replace all quickactivate with quacktivate and pretend like nothing happened. Found Fredrik's comment super funny hahaha

asinghvi17:
`const quacktivate = quickactivate` :D

sebastianpech:
What's your opinion on a short version of `quickactivate(@__DIR__)` like `@quickactivate`? I actually never use anything else than @__DIR__.

tamasgal:
As long as it does not break backwards compatibility, I think that's fine ;) I actually sometimes use relative folders (like `..`)

sebastianpech:
`const 🦆 = quickactivate`

Datseris:
@sebastianpech this is an absolutely mindblowing suggestion. I also only use `@__DIR__`, and your suggestion is totally nonbreaking so all is good. We could also have a 2 argument version of the macro, like
```julia
@quickactivate "project name"
```
which would do the current version does (i.e. check they are the same otherwise error).

sebastianpech:
Good point. You wan't to add it to this PR or should I make an extra one?

Datseris:
meh let's just add it here (but increase to 1.6)

sebastianpech:
hmm it seems i cant push to the project branch

sebastianpech:
`@__DIR__` is a special one ;)

sebastianpech:
I'm currently on the go. Will add doc and bump version later

Datseris:
Good job, thanks!

Datseris:
@sebastianpech sorry, I didn't read your comment about docs and I already did a lot of changes.

You can read and give feedback if I missed something you wanted to do.

sebastianpech:
Thanks, looks good

PR #118, opened 2020-01-29
making usable module docs

Datseris:
Alright, so here I am both writing some useful docs, as well as trying to add a new feature. So, when reading the docs you will see why I find this useful. I make some projects (that are very specific and thus I always use the same packages/etc.) usable modules.

Now, what I am trying to extend is to be able to create a macro so that 
```julia
@quickactivate ProjectName
```
somehow expands to
```julia
quickactivate(@__DIR__, string(ProjectName)
using ProjectName
```
I have a suspicion that this is not possible to do really with `ProjectName`, because this is undefined when we call the macro... BUT, it is certainly possible to do using Symbols, i.e.
```julia
@quickactivate :ProjectName
```
the problem is, I haven't succeeded yet in doing so...  I am testing the current code in my own project by writing 
```julia
using DrWatson
@quickactivate :AlbedoProperties
```
but alas, it errors
```
ERROR: There was an error during testing
ERROR: LoadError: MethodError: no method matching @quickactivate(::LineNumberNode, ::Module, ::QuoteNode)
Closest candidates are:
  @quickactivate(::LineNumberNode, ::Module, ::Symbol) at C:\Users\datse\.julia\dev\DrWatson\src\project_setup.jl:165
  @quickactivate(::LineNumberNode, ::Module, ::String) at C:\Users\datse\.julia\dev\DrWatson\src\project_setup.jl:155
  @quickactivate(::LineNumberNode, ::Module) at C:\Users\datse\.julia\dev\DrWatson\src\project_setup.jl:155
in expression starting at C:\Users\datse\ownCloud\Science Projects\AlbedoProperties\scripts\diff_vs_time.jl:3
```
@sebastianpech you have worked a bit with macros before, perhaps you have some input?

### DIFF
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -114,14 +114,14 @@ it matches the `name`.
 
     **In addition please be very careful to write:**
     ```julia
-    using DrWatson
+    using DrWatson # YES
     quickactivate(@__DIR__)
     using Package1, Package2
     # do stuff
     ```
     **instead of the erroneous:**
     ```julia
-    using DrWatson, Package1, Package2
+    using DrWatson, Package1, Package2 # NO!
     quickactivate(@__DIR__)
     # do stuff
     ```
@@ -148,10 +148,10 @@ end
     @quickactivate
 Equivalent with `quickactivate(@__DIR__)`.
 
-    @quickactivate name
+    @quickactivate name::String
 Equivalent with `quickactivate(@__DIR__, name)`.
 """
-macro quickactivate(name = nothing)
+macro quickactivate(name::String = nothing)
     if __source__.file === nothing
         dir = nothing
     else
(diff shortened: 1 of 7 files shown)


src/project_setup.jl
```diff
@@ -144,33 +144,32 @@ function quickactivate(path, name = nothing)
     return nothing
 end
 
+function get_dir_from_source(source_file)
```
sebastianpech: I just moved the dir retrieval out to not have redundant code.

src/project_setup.jl
```diff
@@ -144,33 +144,32 @@ function quickactivate(path, name = nothing)
...
 Equivalent with `quickactivate(@__DIR__, name)`.
 """
-macro quickactivate(name::String = nothing)
-    if __source__.file === nothing
-        dir = nothing
-    else
-        _dirname = dirname(String(__source__.file))
-        dir = isempty(_dirname) ? pwd() : abspath(_dirname)
-    end
+macro quickactivate(name = nothing)
```
sebastianpech: This must be :Any to still support `@quickactivate`

src/project_setup.jl
```diff
@@ -144,33 +144,32 @@ function quickactivate(path, name = nothing)
...
 end
 
-macro quickactivate(name::Symbol)
-    if __source__.file === nothing
-        dir = nothing
-    else
-        _dirname = dirname(String(__source__.file))
-        dir = isempty(_dirname) ? pwd() : abspath(_dirname)
-    end
+macro quickactivate(name::QuoteNode)
```
sebastianpech: Apparently symbols are captured as `QuoteNode` in macros

sebastianpech:
Good idea. I have similar situations in almost every project.

Datseris:
Thanks a lot man, you really are a macro expert. I knew you would know how to proceed here!

PR #123, opened 2020-02-08
Add `ignore` to `savename`

### LINKED ISSUE #121: Add `ignore` to `savename`
It would be useful to ignore some keys when doing `savename`.

For example

```julia
args = (
    dataset = "mnist",
    n_epochs = 10,
    use_trained = false,
)

savename(args; ignore = [:use_trained]) # => "dataset=mnist_n_epochs=10"
```

xukai92:
Closes #121

### DIFF
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -89,15 +89,16 @@ function savename(prefix::String, c, suffix::String;
         prefix = joinpath(prefix, dpre)
     end
 
-    labels = setdiff(
-        vecstring.((accesses, ignores)) # make it vector of strings
-    )   # ignores overwrites accesses
-    p = sortperm(labels)
+    # Perform access and ignore logic and sort
+    access_labels, ignore_labels = vecstring.((accesses, ignores))  # make it vector of strings
+    label_acc_tuples = collect(zip(access_labels, accesses))        # create tuple
+    filter!(t -> !(t[1] in ignore_labels), label_acc_tuples)        # ignores overwrites accesses
+    sort!(label_acc_tuples; by=(t -> t[1]))                         # sort by labels
+    
     first = prefix == "" || endswith(prefix, PATH_SEPARATOR)
     s = prefix
-    for j ∈ p
-        val = access(c, visibles[j])
-        label = labels[j]
+    for (label, acc) ∈ label_acc_tuples
+        val = access(c, acc)
         t = typeof(val)
         if any(x -> (t <: x), allowedtypes)
             val = roundval(val,digits=digits,scientific=scientific)
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -71,6 +71,7 @@ savename(d, allowedtypes = (String,)) == "mode=double"
 
 rick = (never = "gonna", give = "you", up = "!");
 savename(rick) == "give=you_never=gonna_up=!" # keys are sorted!
+savename(rick; ignores = ["up"]) == "give=you_never=gonna"
 ```
 """
 savename(c; kwargs...) = savename(default_prefix(c), c, ""; kwargs...)
@@ -181,7 +182,7 @@ access(c, key) = getproperty(c, key)
 
 """
     allignore(c)
-Return all the keys `c` will be ignored during [`savename`](@ref).
+Return all the keys `c` that will be ignored in [`savename`](@ref).
 This is an empty tuple by default.
 """
 allignore(c::Any) = ()
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -38,7 +38,7 @@ See also [`parse_savename`](@ref).
   keys `c` can be accessed with, see [`allaccess`](@ref).
 * `ignores = allignore(c)` : You can also specify which specific keys you want
   to ignore with the keyword `ignores`. By default this is an empty 
-  tuple, see [`allignore`](@ref).
+  tuple, see [`allignore`](@ref). Note that `ignores` overwrites `accesses`.
 * `digits = 3` : Floating point values are rounded to `digits`.
   In addition if the following holds:
   ```julia
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -37,7 +37,7 @@ See also [`parse_savename`](@ref).
   to use with the keyword `accesses`. By default this is all possible
   keys `c` can be accessed with, see [`allaccess`](@ref).
 * `ignores = allignore(c)` : You can also specify which specific keys you want
-  to ignore with the keyword `ignores`. By default this is an empty 
+  to ignore with the keyword `ignores`. By default this is an empty
   tuple, see [`allignore`](@ref). Note that `ignores` overwrites `accesses`.
 * `digits = 3` : Floating point values are rounded to `digits`.
   In addition if the following holds:
@@ -47,7 +47,7 @@ See also [`parse_savename`](@ref).
   then the integer value is used in the name instead.
 * `scientific = nothing` : Number of significant digits used for rounding of
   floating point values using scientific notation (e.g. `1.65e-7`).
-  If `nothing`, normal rounding is done. 
+  If `nothing`, normal rounding is done.
 * `connector = "_"` : string used to connect the various entries.
 * `expand::Vector{String} = default_expand(c)` : keys that will be expanded
   to the `savename` of their contents, to allow for nested containers.
@@ -90,15 +90,15 @@ function savename(prefix::String, c, suffix::String;
     end
 
     # Perform access and ignore logic and sort
-    access_labels, ignore_labels = vecstring.((accesses, ignores))  # make it vector of strings
-    label_acc_tuples = collect(zip(access_labels, accesses))        # create tuple
-    filter!(t -> !(t[1] in ignore_labels), label_acc_tuples)        # ignores overwrites accesses
-    sort!(label_acc_tuples; by=(t -> t[1]))                         # sort by labels
-    
+    labels = vecstring(accesses) # make it vector of strings
+    ignored_labels = vecstring(ignores)
+    p = sortperm(labels)
     first = prefix == "" || endswith(prefix, PATH_SEPARATOR)
     s = prefix
-    for (label, acc) ∈ label_acc_tuples
-        val = access(c, acc)
+    for j ∈ p
+        label = labels[j]
+        label ∈ ignored_labels && continue
+        val = access(c, accesses[j])
         t = typeof(val)
         if any(x -> (t <: x), allowedtypes)
             val = roundval(val,digits=digits,scientific=scientific)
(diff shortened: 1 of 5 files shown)


src/naming.jl
```diff
@@ -173,6 +177,13 @@ access(c, keys...) = access(access(c, keys[1]), Base.tail(keys)...)
 access(c::AbstractDict, key) = getindex(c, key)
 access(c, key) = getproperty(c, key)
 
+"""
+    allignore(c)
+Return all the keys `c` will be ignored during [`savename`](@ref).
+This is an empty tuple by default.
+"""
+allignore(c::Any) = ()
```
sebastianpech: I would rename this to `nonignore`. 
`allaccess` mean access all field, then `allignore` means ignore all fields.

sebastianpech: And you don't need to specify `::Any`, it's the default.

src/naming.jl
```diff
@@ -86,12 +89,13 @@ function savename(prefix::String, c, suffix::String;
         prefix = joinpath(prefix, dpre)
     end
 
-    labels = vecstring(accesses) # make it vector of strings
+    visibles = setdiff(accesses, ignores)
+    labels = vecstring(visibles) # make it vector of strings
```
sebastianpech: Currently `allaccess` will return `Vector{String}` or `Vector{Symbol}`. So to make this work, `accesses` and `ignores` must be of the same type. If you switch the lines and convert both with `vecstring` before doing `setdiff` the types don't matter because everything is a string afterwards.

src/naming.jl
```diff
@@ -36,6 +36,9 @@ See also [`parse_savename`](@ref).
 * `accesses = allaccess(c)` : You can also specify which specific keys you want
   to use with the keyword `accesses`. By default this is all possible
   keys `c` can be accessed with, see [`allaccess`](@ref).
+* `ignores = allignore(c)` : You can also specify which specific keys you want
+  to ignore with the keyword `ignores`. By default this is an empty 
+  tuple, see [`allignore`](@ref).
```
sebastianpech: Please add that `ignores` overwrites `accesses`

src/naming.jl
Datseris: Hm, I am in favor of `allignore`. `allaccess` does not mean "access all the fields", but it means "all the fields that will be accessed". It just so happens that by default it _is_ all fields.

Similarly `allignore` means "all the fields that will be ignored". `noignore` isn't so intuitive, because I would naively think that it would return that fields that _wouldn't_ be ignored...

test/naming_tests.jl
```diff
@@ -26,11 +26,13 @@ n = (x = x, y = y)
 
 z = "lala"
 d2 = Dict(:x => x, :y => y, :z => z)
-n2 = (x = x, y = y, z= z)
+n2 = (x = x, y = y, z = z)
 
 @test d2 == @dict x y z
 @test n2 == @ntuple x y z
 
+@test savename(n2; ignores=(:y,)) == "x=3_z=lala"
```
Datseris: A case where both `ingores` and `accesses` are specified must be tested, to ensure that the set difference indeed works as expected.

src/naming.jl
sebastianpech: Ok. I mean it's a function thats not exported anyway so I'm fine with `allignore`

xukai92: I thought the variabels have the same meaning as @Datseris explains. I will keep it as `allignore`.

src/naming.jl
xukai92: Good point! I will fix this.

xukai92: It actually introduces some issue later in the code. Basically the current way of implement the sort is by returning the indics of `visibles` (previously `accesses`) to `p` and latter access it by `visibles[i]`(previously `accesses[i]`). This is necessary because we need the original variables in `visibles` (or `accesses`) for `access`. However, this relies on the fact that the label and value in the label and original container has consistent indics, which is not the case after `setdiff`.

I have to change the way labels and original values are accessed using `zip` and `filter` - please see my updates.

test/naming_tests.jl
xukai92: Added.

src/naming.jl
xukai92: Added.

src/naming.jl
```diff
@@ -89,14 +89,16 @@ function savename(prefix::String, c, suffix::String;
         prefix = joinpath(prefix, dpre)
     end
 
-    visibles = setdiff(accesses, ignores)
-    labels = vecstring(visibles) # make it vector of strings
-    p = sortperm(labels)
+    # Perform access and ignore logic and sort
+    access_labels, ignore_labels = vecstring.((accesses, ignores))  # make it vector of strings
+    label_acc_tuples = collect(zip(access_labels, accesses))        # create tuple
+    filter!(t -> !(t[1] in ignore_labels), label_acc_tuples)        # ignores overwrites accesses
```
Datseris: `setdiff` should be used here instead. Not only it combine these 2 lines into one, it is also more performant.

Datseris:
```

julia> a = ["a", "b", "c"]
3-element Array{String,1}:
 "a"
 "b"
 "c"

julia> b = ["b"]
1-element Array{String,1}:
 "b"

julia> setdiff(a, b)
2-element Array{String,1}:
 "a"
 "c"
```

xukai92: I understand how `setdiff` works. However, what we have now is a list of tuples.
```
julia> a_sym = [("a", :a),  ("b", :b)]; # or

julia> a_str = [("a", "a"), ("b", "b")];

julia> b_sym = [("b", :b)]; # or

julia> b_str = [("b", "b")];

julia> setdiff(a_sym, b_sym)
1-element Array{Tuple{String,Symbol},1}:
 ("a", :a)

julia> setdiff(a_sym, b_str)
2-element Array{Tuple{String,Symbol},1}:
 ("a", :a)
 ("b", :b)

julia> setdiff(a_str, b_sym)
2-element Array{Tuple{String,String},1}:
 ("a", "a")
 ("b", "b")

julia> setdiff(a_str, b_str)
1-element Array{Tuple{String,String},1}:
 ("a", "a")
```

The current implementation can deal with both cases.

Besides, I think performance should not be a concern here.

Datseris: I must admit I am extremely confused by this comment. Why would there be a datastructure like `a_sym = [("a", :a),  ("b", :b)];`???

`allaccess` and `allignore` both return a vector of keys to access the container. These are either vectors of symbols or vectors of strings, or the empty tuple `()`. In any case `setdiff` will always return the same, simple type as `accesses`.

I now see you have increased the complexity of the source without reason. Unless I am missing something, what you have to do is simply:
```julia
accesses = setdiff(accesses, ingores) # type stable operation, even if `ignores = ()`
```
just before the line `labels = vecstring(accesses)`. Then _change nothing of the subsequent code_. @sebastianpech am I missing something here? Because I think its that simple.

> Besides, I think performance should not be a concern here.

Well, on one hand we have a version that does (a) only a single line modification of the function's core, (b) is more performant and (c) is simpler. The best version is the version that should be implemented. Wouldn't you agree?

src/naming.jl
xukai92: @Datseris Please see here on why simply swapping the lines doesnt work.

src/naming.jl
xukai92: Please read [my comment on the other conversation](https://github.com/JuliaDynamics/DrWatson.jl/pull/123#discussion_r376717544) on why simply swapping the lines as suggested by @sebastianpech doesn't work.

Datseris:
Okay, @xukai92 I see. I have been proven wrong, and it wasn't as simple as I initially thought. 

However, I am not yet giving up in trying to find ways to make this simpler. After all, DrWatson's goal of existence is to make things simple ;)

Here is another suggestion, which modifies the original code as follows. First, define
```julia
    ignore_labels = vecstring(ignores)
```
as you do, and then don't do anything extra. Then, in the main loop of the original code:
```julia
for j ∈ p
        val = access(c, accesses[j])
```
add the line line
```julia
labels[j] in ignore_labels && continue
```
which would skip this iteration.

Would this work just as well? Or there would need modifications to be made? At a first glance, I think this would "work", but I've been wrong before as you know already.

src/naming.jl
xukai92: It might be hard to understand it by words. Here is what would happen if I simply change the two lines.

```julia
    labels_acc, labels_ign = vecstring(accesses), vecstring(ignores)  # L1
    labels = setdiff(labels_acc, labels_ign)                          # L2
    p = sortperm(labels)                                              # L3
    first = prefix == "" || endswith(prefix, PATH_SEPARATOR)          # L4
    s = prefix                                                        # L5
    for j ∈ p                                                         # L6
        val = access(c, accesses[j])                                  # L7
        label = labels[j]                                             # L8
```

Given `accesses = [:z, :y, :x]` and `ignores = [:z]`
- L1: `labels_acc <- ["z", "y", "x"]` and `labels_ign <- ["z"]`
- L2: `labels <- ["y", "x"]`
- L3: `p <- [2, 1]`
- Loop (first time)
  - L6: `j <- 2`
  - L7: `accesses[j]` is `:y`
  - L8: `labels[j]` is `"x"`
- Loop (second time)
  - L6: `j <- 1`
  - L7: `accesses[j]` is `:z`
  - L8: `labels[j]` is `"y"`

As you can see, `accesses[j]` is not consistent with `labels[j]`.

Datseris: Yes, I have understood that, which is why in my latest comment I suggest the `continue` version. When I say "original" version of the code, I literally mean the version before your PR, here: https://github.com/JuliaDynamics/DrWatson.jl/blob/master/src/naming.jl#L89-L94

Adding a `labels[j] in ignore_labels && continue` would fix everything. See https://github.com/JuliaDynamics/DrWatson.jl/pull/123#issuecomment-583779631 . Please notice what I suggest: there is no `labels_acc, labels_ign, labels`, there is only `labels` and just some iteration steps are skipped.

Datseris:
@xukai92 I've just implemented it locally. Adding `    ingored_labels = vecstring(ignores)` and the command 
```
label ∈ ingored_labels && continue
```
in the loop achieves what we want. Full code:
```julia
    labels = vecstring(accesses) # make it vector of strings
    ingored_labels = vecstring(ignores)
    p = sortperm(labels)
    first = prefix == "" || endswith(prefix, PATH_SEPARATOR)
    s = prefix
    for j ∈ p
        label = labels[j]
        label ∈ ingored_labels && continue
        val = access(c, accesses[j])
```

which for example does
```
julia> savename(a)
"x=3_y=2_z=5"

julia> savename(a; ignores = ["z"])
"x=3_y=2"
```

xukai92:
Thanks @Datseris ! Your `continue` solution indeed looks much nicer than my filtering one.

Datseris:
You're welcome, thanks for the idea and the contribution. I am always going full crazy mode regarding code clarity, I hope I didn't annoy you _that_ much. :D

xukai92:
That's all right! I'm happy that I learned how to use `continue` better :p

sebastianpech:
I get it now. The `continue` solution is a lot easier to read. Thanks for your work

PR #157, opened 2020-05-25
Support keyword arguments in `wsave` and `wload`

devmotion:
I noticed that it's currently not possible to use `wsave` with keyword arguments (needed, e.g., when setting the output resolution in CairoMakie: https://github.com/JuliaPlots/CairoMakie.jl#saving).


Datseris:
Thank you @devmotion , would you mind incrementing the minor version in Project.toml as well?

PR #165, opened 2020-06-29
Make savename work with `Inf` and `NaN` values

JonasIsensee:
Currently `savename` errors when you try to use it with `Inf` or `NaN` floats.
(The issue happens when trying to round to a certain number of digits)


Datseris:
Cool, can you please increment patch version in Project.toml?

PR #168, opened 2020-07-06
Allow additional syntax for collect results

JonasIsensee:
This PR implements an additional syntax for the `collect_results` function.

It enhances the `special_list` field by allowing two different notations.

```
special_list = [
   :vl_mean => data -> mean(data["c"]),
   
  data -> [:field1 => rand() , :field2 => rand()],
]
```


src/result_collection.jl
```diff
@@ -90,7 +96,7 @@ function collect_results!(filename, folder;
     end
 
     n = 0 # new entries added
-    existing_files = "path" in names(df) ? df[:,:path] : ()
+    existing_files = "path" in string.(names(df)) ? df[:,:path] : ()
```
JonasIsensee, line 99: The behaviour that `names(df)` returns a Vector of strings is only very recent.
I have now encountered this multiple times that DataFrames is restricted to something older
and suddenly DrWatson fails.

Datseris: Yeap the dataframes guys did this breaking change that now `names` return vector of strings. we have to update DrWatson in general to the latest version....

src/result_collection.jl
```diff
@@ -43,9 +43,15 @@ See also [`collect_results`](@ref).
...
-is a derived quantity to be included in `df`. The function entry always
+`special_list` is a `Vector` where each entry
+is a derived quantity to be included in `df`. There are two types of entries.
+The first option is of the form `key => func` where the `key` is a symbol
+to be used as column name in the DataFrame. The function entry always
 takes a single argument, which is the loaded result-file (a dictionary).
+The second option is to provide just one function `func`. This function
+also takes the single dictionary argument but returns one or more
+`key => value` pairs. This second notation may be useful when a single
+when one wants to extract values for multiple columns in a single step.
```
Datseris, line 54: The last sentence here doesn't make sense (grammatically).

JonasIsensee: fixed it

src/result_collection.jl
JonasIsensee: yeah, I figured that this wouldn't hurt.
Converting strings to strings shouldn't be too expensive.
And this way we at least know that this will work without having to put a new lower version bound on DataFrames.

PR #169, opened 2020-07-07
Add feature for allowing parameter restriction for dicts passed to dict_list

sebastianpech:
**TLDR**
New macro `@onlyif` that allows placing restrictions on certain parameters in a Dict.
`dict_list` then generates the parameters according to those restrictions.

```julia
p = Dict(
         :α => 1,
         :solver => ["SolverA","SolverB"],
         :c => @onlyif(:solver == "SolverA", [100,200]),
         :d => @onlyif(:solver == "SolverB", 1)
        )
dict_list(p)
```
Gives:

```
 3-element Array{Dict{Symbol,Any},1}:
  Dict(:α => 1,:solver => "SolverA",:c => 200)
  Dict(:α => 1,:d => 1,:solver => "SolverB")
  Dict(:α => 1,:solver => "SolverA",:c => 100)
```

**Details**
I really like using `dict_list` because it's such a neat function with a very simple concept.
On thing that bothered me lately was that I often have cases where I wanted to do "parameter sweeps" on parameters that only occur if another parameter is set to a specific value.
So for example I have a parameter `:solver` which specifies a solver type for the simulation and those types specified by `:solver` have different parameters (In the following example `:c` for SolverA and `:d` for SolverB).
One way to define this is:

```julia
p = Dict(
         :α => 1,
         :solver => ["SolverA","SolverB"],
         :c => [100,200],
         :d =>  1
        )
```
However, `dict_list` now gives me a bunch of parameter combination which I actually don't need. (Like everything with `:solver = "SolverA"` that contains `:d = 1`.

I think I found a nice solution for defining restrictions or dependencies on the parameters. I wanted to stick as closely as possible to the above notation so I came up with this:

```julia
p = Dict(
         :α => 1,
         :solver => ["SolverA","SolverB"],
         :c => @onlyif(:solver == "SolverA", [100,200]),
         :d => @onlyif(:solver == "SolverB", 1)
        )
```

The first argument of `@onlyif` is a Julia expression defining the condition for this parameter and the second argument are the values for the parameter.
This returns this Dict:

```
Dict{Symbol,Any} with 4 entries:
   :α      => 1
   :d      => DependentParameter{Int64}(1, #22)
   :solver => ["SolverA", "SolverB"]
   :c      => DrWatson.DependentParameter{Int64}[DependentParameter{Int64}(100, #21), DependentParameter{Int64}(200, #21)]
```

As you can see, the restricted parameters are now wrapped in a new type that contains the value and a function defining if the parameter is valid for a given trial combination (trial combination = A parameter combination that not necessarily matches the constraints).
The macro is actually quite simple and performs the following steps:

1. Replace every `QuotedNode` and `String` with a function call, such that it gives the value of the referenced parameter. (The function `lookup_candidate` does this)
2. Wrap the adapted expression into an anonymous function taking a trial parameter combination as only argument

For the above line for parameter `:d`, `@onlyif` produces the following code:
```julia
:(DrWatson.toDependentParameter(1, (pd->begin
              DrWatson.lookup_candidate(pd, :solver) == DrWatson.lookup_candidate(pd, "SolverB")
          end)))
```
Here, `pd` is the trial combination you would also get when calling the original `dict_list` function. `DrWatson.lookup_candidate(pd, :solver)` returns `pd[:solver]` if  `:solver` is a key of `pd` or `:solver` if not.
So in this case
```
DrWatson.lookup_candidate(pd, :solver) == DrWatson.lookup_candidate(pd, "SolverB")
```
is equal to 
```
pd[:solver] == "SolverB"
```

To get the final list, I generate all possible parameter combinations and then filter them such that no parameter combinations contains any restricted parameter that violates its assigned function.
The `Set` removes any duplicates.

Calling `dict_list` with the above dict now gives:
```
 3-element Array{Dict{Symbol,Any},1}:
  Dict(:α => 1,:solver => "SolverA",:c => 200)
  Dict(:α => 1,:d => 1,:solver => "SolverB")
  Dict(:α => 1,:solver => "SolverA",:c => 100)
```


Datseris:
This is **seriously awesome**!

1. Increment minor version
2. Add documentation/example somewhere
3. Add changelog

and we ship it!

sebastianpech:
That's nice to hear :) ... One info about MacroTools.jl ..... it's not a v1.
However, it's been around for quite a while already and 510 other packages depend on it (https://juliaobserver.com/packages?dependent_id=MacroTools)

Datseris:
God damn it.

But why do you need the whole package? Seems to me that by checking the source, what you need is:
```julia
walk(x, inner, outer) = outer(x)
walk(x::Expr, inner, outer) = outer(Expr(x.head, map(inner, x.args)...))
postwalk(f, x) = walk(x, x -> postwalk(f, x), f)
```

am I incorrect?

JonasIsensee:
This looks super cool!

sebastianpech:
I copied the code from MacroTools over. I will test the feature on a more advanced example tomorrow. If everything works I add the docs and finish the rest.

sebastianpech:
@Datseris @JonasIsensee can you please check if the docstrings I added make any sense to you (https://github.com/JuliaDynamics/DrWatson.jl/pull/169/commits/742f84a56c3f4d96f686066b3ed9773efebe97b7)

src/dict_list.jl
```diff
@@ -166,13 +166,14 @@ the Julia expression `ex` is evaluated as true.  If `value` is a subtype of
 `Vector`, `@onlyif` is applied to each entry.
 Since '@onlyif' is applied to a value and not to a dictionary key, it is
 possible to restrict only some of the values of a vector. This means that based
-on on 'ex' the number of options for a particular key varies.
+on on `ex` the number of options for a particular key varies.
```
Datseris, line 169:
```suggestion
on `ex` the number of options for a particular key varies.
```

src/dict_list.jl
```diff
@@ -0,0 +1,210 @@
...
+        end
+        return d[name]
+    end
+    return name
+end
+
+"""
+    @onlyif(ex, value)
+Tag `value` to only appear in a dictionary created with [`dict_list`](@ref) if
+the Julia expression `ex` is evaluated as true.  If `value` is a subtype of
```
Datseris, line 165: Hm, here I was confused about what `ex` is.. I thought it would be a boolean value, but it is actually a Julia expresision containing previous keys of the `dict_list` argument, represented as symbols, and operations involving them. Perhaps we should say this.

Datseris: The examples of course clarify this as well.

Datseris: Ah, forget this whole coversation, I see you explain it perfectly below.

Datseris, line 165:
```suggestion
the Julia expression `ex` (see below) is evaluated as true.  If `value` is a subtype of
```

src/dict_list.jl
```diff
@@ -0,0 +1,210 @@
...
+provided.  For example `ex = :(:N == 1)` is tranformed in the call
+`dict_list(d)` to an expression analogous to `:(d[:N] == 1)` by using the
+function [`lookup_candidate`](@ref).  This is supported for `Symbol` and
+`String` keys.
+
+## Examples
+```julia
+julia> c = Dict(:a => [1, 2], :b => 4, :c => @onlyif(:a == 1, [10, 11]));
+
+julia> dict_list(c)
```
Datseris, line 182:
```suggestion
julia> dict_list(c) # only the case of `a=1` will get key `c`
```

src/dict_list.jl
```diff
@@ -0,0 +1,210 @@
...
+
+julia> dict_list(c) # only the case of `a=1` will get key `c`
+3-element Array{Dict{Symbol,Int64},1}:
+ Dict(:a => 1,:b => 4,:c => 10)
+ Dict(:a => 1,:b => 4,:c => 11)
+ Dict(:a => 2,:b => 4)
+
+ julia> c = Dict(:a => [1, 2], :b => 4, :c => [10, @onlyif(:a == 1, 11)]);
+
+julia> dict_list(c)
```
Datseris, line 190:
```suggestion
julia> dict_list(c) # case of `a=1` will get extra value `11` for key `c`
```

sebastianpech:
I'll try to make a nice and simple real world example. The one I have is a bit too complicated I guess

src/dict_list.jl
```diff
@@ -177,17 +177,17 @@ function [`lookup_candidate`](@ref).  This is supported for `Symbol` and
 
 ## Examples
 ```julia
-julia> c = Dict(:a => [1, 2], :b => 4, :c => @onlyif(:a == 1, [10, 11]));
+julia> d = Dict(:a => [1, 2], :b => 4, :c => @onlyif(:a == 1, [10, 11]));
 
-julia> dict_list(c) # only the case of `a=1` will get key `c`
+julia> dict_list(d) # only the case of `:a == 1` will get key `:c`
```
Datseris, line 182: I am not sure that using `:a == 1` is the best description for a comment. `:a == 1` is equivalent to `false` in Julia. In reality what we want to say is that "when the key `:a` is `1`, then we get key `:c`.

sebastianpech: Sure, good point.

src/dict_list.jl
```diff
@@ -0,0 +1,210 @@
...
+function dict_list(c::Dict)
+    if contains_partially_restricted(c) || contains_fully_restricted(c)
+        # The method for generating the restricted parameter set is as follows:
+        # 1. Create an array of trial combinations containing all possible
+        # combinations of all parameters.
+        # 2. For each solution, remove all parameters where the conditions
+        # functions returns false
+        # 3. Do (2) until the length of the obtained parameter set stops
+        # changing.
+        # 4. Replace all `DependendParameter` types by their respective values
```
JonasIsensee, line 59:
```suggestion
        # 4. Replace all `DependentParameter` types by their respective values
```

docs/src/real_world.md
```diff
@@ -397,3 +397,25 @@ res = collect_results!(datadir("results"); black_list = bl, subfolders = true)
...
+```julia
+ga_parameters = Dict(
+    :population_size => [20,50,100],
+    :selection => ["roulette-selection", "SUS", "tournament-selection", "linear ranking"],
+    :fitness_scaling => @onlyif(:selection in ("SUS", "roulette-selection"), collect(1.0:20.0)),
+    :tournamet_size => @onlyif(:selection == "tournament-selection", collect(2:10)),
+    :chromosome => [:A, @onlyif(begin
+        size_constr = :population_size <= 50
+        select_constr = :selection != "SUS"
+        size_constr && select_constr
```
Datseris, line 417: `size_constr && select_constr` what does this line do?

Datseris: AHA! It says "only if condition `size_constr` and `select_constr` are met! Maybe you add a comment to this after the exectution for clarity

Datseris: Damn, gotta say, this is so pretty advanced stuff :D

sebastianpech:
I added a condensed example from my master thesis and also showed that one can use `begin end` blocks.
@Datseris is it possible to preview the generated docs?

docs/src/real_world.md
sebastianpech: Usually I would put this on one line, but I wanted to show, that one can use `begin end`.
Both variables are boolean

JonasIsensee: only if both conditions are met, return true?

Datseris:
> @Datseris is it possible to preview the generated docs?

It would be if you had a branch instead of a fork :D In the next PR branch directly from the origin instead of the fork and then it will actually happen automatically.

sebastianpech:
Oh, I didn't know I can do that :)

docs/src/real_world.md
sebastianpech: yeah right. One could reason (to have a real world example), that this type of chromosome has a bad performance for a population size larger than 50 and one needs to use a fast selection algorithm ...

Datseris:
https://github.com/orgs/JuliaDynamics/teams/drwatson/members

docs/src/real_world.md
```diff
@@ -419,3 +419,7 @@ ga_parameters = Dict(
 
 dicts = dict_list(ga_parameters)
 ```
+
+The parameter restriction for the chromosome type shows that one can use arbitrary Julia expressions that return `true` or `false`.
+In this case, first the conditions for the population size and for the selection method are evaluated and stored.
+The expression then only returns true, if both conditions are met, thus restricting the usage of chromosome type `:B`.
```
sebastianpech, line 425: I added a bit of extra info about the advanced part

docs/src/real_world.md
```diff
@@ -397,3 +397,29 @@ res = collect_results!(datadir("results"); black_list = bl, subfolders = true)
...
+
+```julia
+ga_parameters = Dict(
+    :population_size => [20,50,100],
+    :selection => ["roulette-selection", "SUS", "tournament-selection", "linear ranking"],
+    :fitness_scaling => @onlyif(:selection in ("SUS", "roulette-selection"), collect(1.0:20.0)),
+    :tournamet_size => @onlyif(:selection == "tournament-selection", collect(2:10)),
+    :chromosome => [:A, @onlyif(begin
+        size_constr = :population_size <= 50
+        select_constr = :selection != "SUS"
```
Datseris, line 416:
```suggestion
        size_constr = (:population_size <= 50)
        select_constr = (:selection != "SUS")
```

I suggest that this will increase clarity, because typically `varname = :symbol` is interpreted as assignement at first glance.

docs/src/real_world.md
Datseris: Awesome

src/dict_list.jl
```diff
@@ -0,0 +1,210 @@
...
+ Dict(:a => 2,:b => 4)
+
+ julia> d = Dict(:a => [1, 2], :b => 4, :c => [10, @onlyif(:a == 1, 11)]);
+
+julia> dict_list(d) # only in case `:a` is `1` the dictionary will get extra value `11` for key `:c`
+3-element Array{Dict{Symbol,Int64},1}:
+ Dict(:a => 1,:b => 4,:c => 10)
+ Dict(:a => 1,:b => 4,:c => 11)
+ Dict(:a => 2,:b => 4,:c => 10)
+```
```
Datseris, line 195:
```suggestion
```

See the [Defining parameter sets with restrictions](@ref) section for more examples.

```

docs/src/real_world.md
sebastianpech: Really? I don't know. I'm pretty used to the fact that `=` has the lowest operator precedence.

Datseris:
Hm, unfortunately the example did not build:
https://juliadynamics.github.io/DrWatson.jl/dev/real_world/#Defining-parameter-sets-with-restrictions-1


which is odd because there is no error in the build log https://travis-ci.org/github/JuliaDynamics/DrWatson.jl/jobs/706304388#L283

sebastianpech:
@Datseris I did't mark the code block as `@example customizing`. You think that's useful? The code spits out ~ 700 combinations.

Datseris:
aaaaaah okay!

PR #170, opened 2020-07-09
Create project structures that can be cloned as is

UtkarshChemE:
Added an optional argument `placeholder` to enable cloning of the project structure. It simply adds a hidden file `.placeholder` in all folders and commits it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.14.1"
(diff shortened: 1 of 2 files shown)


Datseris:
Thanks,

1. you need to add `placeholder` in the documentation string
2. also increase patch version in the `Project.toml` file (as this is mostly a "fix" than a new feature)

and then we can merge :)

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.14.0"
+version = "1.14.1"
```
```suggestion
version = "1.14.2"
```

PR #171, opened 2020-07-09
Update real world example "Defining parameter sets with restrictions"

sebastianpech:


sebastianpech:
I think I just stumbled upon a bug

Datseris:
What bug?

Datseris:
HEre is the built docs by the way: https://juliadynamics.github.io/DrWatson.jl/previews/PR171/real_world/#Defining-parameter-sets-with-restrictions-1

sebastianpech:
> What bug?

Well a case where I can't deduct if the key is fully restricted or not just by looking at the value's type.
When every value of that key in the dictionary is restricted, but based on the conditions the key is never fully excluded.

```
p = Dict(
         :a => [1,2],
         :b => @onlyif(:a==1,[1,2]),
         :c => [@onlyif(:b==1,1), @onlyif(:b==2,2)],
        )
```

gives 
```
julia> dict_list(p)
5-element Array{Dict{Symbol,Int64},1}:
 Dict(:a => 1,:b => 1,:c => 1)
 Dict(:a => 1,:b => 2,:c => 2)
 Dict(:a => 2)
 Dict(:a => 1,:b => 1)
 Dict(:a => 1,:b => 2)
```

Last two shouldn't be there, because `:c` is valid for all `:b`.
I'm working on it

sebastianpech:
That took me far too long to find that the solution is actually quite simple 🤯.
This approach makes some of the code even simpler now. So my approach is as follows:
Looking at 
```
 Dict(:a => 1,:b => 1,:c => 1)
 Dict(:a => 1,:b => 2,:c => 2)
 Dict(:a => 2)
 Dict(:a => 1,:b => 1)
 Dict(:a => 1,:b => 2)
```
It obviously makes not sense to include 
```
 Dict(:a => 1,:b => 1)
 Dict(:a => 1,:b => 2)
```
in the solution, because these parameter settings are already present in
```
 Dict(:a => 1,:b => 1,:c => 1)
 Dict(:a => 1,:b => 2,:c => 2)
```
If someone wants this, they would need to explicitly do smth like this:
```
 :c => [@onlyif(:b in (1,2), nothing), @onlyif(:b==1,1), @onlyif(:b==2,2)]
```
to get
```
 Dict(:a => 1,:b => 1, :c=>nothing)
 Dict(:a => 1,:b => 2, :c=>nothing)
```
in order to use their default values for `:c` or whatever

PR #172, opened 2020-07-09
Extended `mkpath` to handle placeholders.

UtkarshChemE:


Datseris:
Hi @UtkarshChemE , I'm sorry but we can't do what you suggest in this PR. This is a case of **type piracy**. It occurs when you are extending a method where both of the following is *not* true:

1. You own the function name.
2. You own at least one of the types.

This in general is not considered code hygienic . Why do we need this extention instead of a new function?

UtkarshChemE:
Well I don't know what is the best way to go forward. I think creating a function with unique name might be fine. I will update my code to reflect the changes.

Datseris:
Okay, so what is the point of `mkpath_with_placeholder` ? Is it used somewhere?

UtkarshChemE:
`mkpath_with_placeholder` is a convenience utility. Once, the project is created and one would like to add additional directories to manage their data or plots, they can use `mkpath_with_placeholder`. This will ensure that if they clone their project, elsewhere the directory structure would be maintained, without breaking their automated scripts.

sebastianpech:
@UtkarshChemE why would you create a directory after the project was initialized without putting anything in it?

UtkarshChemE:
Check out [NNPublic](https://github.com/UtkarshChemE/NNPublic). I have scripts that expect files in particular sub directory of `data`. They read and write data in that sub-directories and I would like to retain that structure when I clone my project on another machine.

Datseris:
Have you seen the `wsave` function, which always creates the necessary path before saving a file? This may eliminate your problem of not having folders in your computer? (I am trying to imagine in what other scenario you absolutely need the existence of an empty folder).

UtkarshChemE:
The only time I need empty folders available is when I am reading data from that folders. I have a complex mess of multi-projects and they talk to each other in an automated fashion for sequential processing. Maybe adding a function to create blank commit-able folders is excessive. I think this function is better suited in my `startup.jl`.

Datseris:
The Julia function `touch` is what you want I think.

PR #174, opened 2020-07-13
Update project_setup.jl

grahamstark:
setup failed if `placeholder` = true; I think `files` should be `ph_files` here.


Datseris:
Yeap, thanks. Please increment patch version in Project.toml and we can merge this :)

PR #176, opened 2020-07-15
cleaned some typos in the docs

koehlerson:
Hey there,

today when I went through the `Running & Listing Simulations` I noticed some typos that I removed. 
The `lookup_candidate` ref isn't really a typo but since the docs don't include this function, the ref was directing to 404.

Cheers
Max


sebastianpech:
@koehlerson cool thanks. I wasn't sure what happens if an entry has no reference in the docs but forgot to check if the ref is working.

PR #179, opened 2020-07-20
Add DOI for Sumatra

tamasgal:
...sorry, the DOI of the original citation for Sumatra which I copy&pasted was missing, I had to add it manually...


tamasgal:
Thanks!

Issue #116, opened 2020-01-19
collect_results! : path


JonasIsensee:
When collecting results using `collect_results!` the path to the files is automatically added.
This is great but we currently add the full system-path.
For better portability one could change this to just save the path relative to the project-rootfolder.
Then things would stay coherent over different machines.

Datseris:
Such action is already implemented in `@tagsave` and company, at the point where we store the path to the script: see `function tag!` here https://github.com/JuliaDynamics/DrWatson.jl/blob/master/src/saving_tools.jl#L173

so I guess we just need `relpath(path, projectdir)`

PR #180, opened 2020-07-22
Use relpath in collect_results!

nmheim:
Addresses #116 

Removes `datadir()` from the path entry before adding a row to the collected results dataframe. I can change it to remove only `projectdir` instead, using the data directory just seemed more natural to me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 1.14.5
+* new keyword argument `rpath` for `collect_results!` that allows storing relative paths
 # 1.14.0
 * New macro `@onlyif` that allows placing restrictions on values in a dictionary used for expansion with `dict_list`
 # 1.13.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 1.14.5
+# 1.14.6
 * new keyword argument `rpath` for `collect_results!` that allows storing relative paths
 # 1.14.0
 * New macro `@onlyif` that allows placing restrictions on values in a dictionary used for expansion with `dict_list`
--- a/src/result_collection.jl
+++ b/src/result_collection.jl
@@ -36,6 +36,8 @@ See also [`collect_results`](@ref).
   for result-files.
 * `valid_filetypes = [".bson", ".jld", ".jld2"]`: Only files that have these
   endings are interpreted as result-files. Other files are skipped.
+* `rpath = nothing` : If not `nothing` stores `relpath(file,rpath)` of result-files
+  in `df`. By default the absolute path is used.
 * `verbose = true` : Print (using `@info`) information about the process.
 * `white_list` : List of keys to use from result file. By default
   uses all keys from all loaded result-files.
--- a/src/result_collection.jl
+++ b/src/result_collection.jl
@@ -103,11 +103,11 @@ function collect_results!(filename, folder;
     for file ∈ allfiles
         is_valid_file(file, valid_filetypes) || continue
         # maybe use relative path
-        file = isnothing(rpath) ? file : relpath(file, rpath)
+        file = rpath === nothing ? file : relpath(file, rpath)
         #already added?
         file ∈ existing_files && continue
 
-        data = isnothing(rpath) ? wload(file) : wload(joinpath(rpath, file))
+        data = rpath === nothing ? wload(file) : wload(joinpath(rpath, file))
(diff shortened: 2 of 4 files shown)


Datseris:
Thanks a lot. I would personally still vote to have the `projectdir` as the relative level:

1. For consistency with other functions that also use relative paths
2. Because it is not really necessary for "data" to be in `data` in the sense that some users (like me) typically create extra directories which may also contain things that can be collected.

src/result_collection.jl
```diff
@@ -88,7 +88,7 @@ function collect_results!(filename, folder;
         allfiles = String[]
         for (root, dirs, files) in walkdir(folder)
             for file in files
-                push!(allfiles, joinpath(root,file))
+                push!(allfiles, relpath(joinpath(root,file), datadir()))
```
Datseris, line 91: it might be worth it to make the relative directory a keyword argument, so that you have `rpath = projectdir()`, and use `rpath` in `relpath`.

Datseris: But I am not sure... At least that's how we do it in `tag!` where `gitpath` is given as argument., This way, if somneone really wants the absolute path, they can provide `""` as `rpath`.

JonasIsensee:
This needs some thought.

I fear that this is not general enough and I'm pretty sure it will break my workflow.
I work in a setup where all data raw and processed lies on different volumes.
(e.g. raw data on /scratch and processed on /data)
This is necessary due to the sheer amount of data and because data files are shared with my colleagues.

Therefore my data definitely does not lie within my `projectdir` and I heavily rely on the absolute file paths.
What I'm trying to say: What you have implemented looks good and can possibly be a reasonable default behaviour but there must be a way/ arg to retain the current behaviour.

nmheim:
Yes, of course that makes a lot of sense. Would a keyword argument along the lines of Datseris suggestion be ok? If so, should the default be to use the absolute path (to avoid breaking code) or the projectdir?

Datseris:
I agree with this. The default could be simply `nothing`, in which case absolute path is used, and otherwise be `projectdir()` or `datadir()`, in which case the relative with the given argument is saved. For backwards compatibility, the default has to be kept as `nothing`.

nmheim:
I added the keyword argument. In case you don't like the if statements with `isnothing` we could probably also do something like this:
```
maybe_relpath(file::String, rpath::Nothing) = file
maybe_relpath(file::String, rpath::String) = relpath(file, rpath)
```
Once we agree I can add an entry to the docstring as well :)

src/result_collection.jl
```diff
@@ -88,7 +89,7 @@ function collect_results!(filename, folder;
         allfiles = String[]
         for (root, dirs, files) in walkdir(folder)
             for file in files
-                push!(allfiles, relpath(joinpath(root,file), datadir()))
+                push!(allfiles, joinpath(root,file))
```
Datseris: Shouldn't there be a `isnothing(rpath)` clause here?

Datseris: And if true use `relpath(joinpath(root, file), rpath))` ?

nmheim: as it is written now all the checks for relative/absolute paths happen in the second loop: https://github.com/JuliaDynamics/DrWatson.jl/blob/ed6cee2110686c771d063923e9f093a77b25cd37/src/result_collection.jl#L104-L108

which is why I suggested using something like `maybe_relpath` (and then also `maybe_joinpath`), but I thought having it all in one loop is easier to read...

Datseris: Sorry, you are correct, I have simply not seen the entire code :)

nmheim:
Why do some of the tests fail with `isnothing` is not defined?

Datseris:
`isnothing` is not defined in Julia 1.0, only in later versions. Just use `rpath === nothing` instead.

Datseris:
@JonasIsensee feel free to merge once you also have a look.

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# 1.14.5
```
Datseris, line 1: Wait, the versions don't match right? I see below 1.14.6 in the Project file.

nmheim: ohhh, sorry! thanks for catching that.

nmheim:
Thank you for these quick replies guys! :)

JonasIsensee:
Looks good to me.
Thank you for your contribution, @nmheim!

PR #185, opened 2020-08-14
Better README in initialize_project

Datseris:
That highlights reproducibility and instructs how to do it.


sebastianpech:
Cool idea, I like it.

PR #189, opened 2020-09-15
add Pluto.jl to notebooks packages

kescobo:


sebastianpech:
Is it actually working with DrWatson? First time I tried I couldn't activate the project.

kescobo:
Ahh, good question, I did not check. I actually haven't used Pluto yet myself, but it seems to be the new hotness. If it's not working, would be nice to look into it, but OK with me to close this pending that investigation.

sebastianpech:
I will check today. I think they recently did some work on supporting environments.

sebastianpech:
Yep. Works now.

PR #192, opened 2020-09-15
Paper review

kescobo:
Related to [JOSS review](https://github.com/openjournals/joss-reviews/issues/2673#event-3765935810)

I've made a few minor edits to the manuscript. A couple of additional points:

- I think it would also be good to mention in the `Comparison with existing software section`, where you discuss that many projects use multiple programming languages, that julia actually has really good interop with other languages. via eg PyCall and RCall.
- In the same section, I think cwl is orthogonal to the purpose of DrWatson. The idea with CWL and other pipeline-type software (see [this discourse](https://discourse.julialang.org/t/rfc-mr-phelps-a-distributed-workflow-orchestrator/) thread for a bunch of other examples) is the run the same analysis pipeline on lots of different datasets. There's obviously some parallels with DrWatson in terms of reproducibility and project management, but they seem like overall they're targeting different concerns.
- Prior to lines 134/135 where you have commands from the julia pkg REPL, it might be worth mentioning that, and providing a link to Pkg documentation. The other code in the manuscript is intended to be put inside julia scripts, and non-julia users are unlikely to be familiar with the `pkg>` prompt convention.
- Similarly, the output in ln115-121 is not clearly distinct from the code starting at ln109. This may be a limitation of the document rendering engine, but it might be clearer if you write something like:

````

```julia
@tagsave(savename(parameters, "jld2"), Dict("data" => [1,2,3]));

load(savename(parameters, "jld2")) # load back saved data
```

yielding the output:

```
Dict{Symbol,Any} with 6 entries:
  :gitcommit => "v1.13.0-1-g3a5364f"
  :script    => "docs/build/string#3"
  :data      => [1, 2, 3]
  :gitpatch  => ""
```
````

I have a few additional thoughts for features that would be useful, but in my mind they are not essential and the feature set currently available is very useful and more than sufficient for publication.

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -131,8 +131,9 @@ The function `collect_results` can traverse the data folder and collect all save
 Sharing and reproducing a DrWatson project is in every respect trivial.
 The entire project folder is simply sent to a different machine, and in a Julia session the user does the following:
 ```
-pkg> activate path/to/project
-pkg> instantiate
+using Pkg
+Pkg.activate("path/to/project")
+Pkg.instantiate()
 ```
 and all necessary dependencies are installed automatically.
 Since the project uses only relative paths because of the function `projectdir`, every script runs as it did on the original machine.
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -73,7 +73,7 @@ In this section we demonstrate how using DrWatson makes the typical scientific w
 This section is a brief summary of the [DrWatson Workflow Tutorial](https://juliadynamics.github.io/DrWatson.jl/dev/workflow/), which in itself showcases a subset of DrWatson's functionality.
 There the workflow is discussed and demonstrated more thoroughly via explicitly running every code command.
 
-Typically one starts a scientific project with the function `initialize_project`.
+Typically, one starts a scientific project with the function `initialize_project`.
 This creates a project folder that contains sensible default structure (e.g. folders for data, papers, scripts, etc.), while also making the project a Julia project.
 This allows the scientific project to be tied with the full hierarchy of exact package versions used, which remains entirely independent from the main Julia installation (or any other project).
 The project is also a git repository, which allows code versioning and reproducibility, and DrWatson provides functions that make this process seamless (see below).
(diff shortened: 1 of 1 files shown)


tamasgal (approved):
Thanks!

Datseris:
We should implement all suggestions in this PR and then merge.

tamasgal:
Yes of course, I’m on it, but I need one or two days.

jpfairbanks:
I generally agree with @kescobo, but I think that 

> In the same section, I think cwl is orthogonal to the purpose of DrWatson. The idea with CWL and other pipeline-type software (see this discourse thread for a bunch of other examples) is the run the same analysis pipeline on lots of different datasets. There's obviously some parallels with DrWatson in terms of reproducibility and project management, but they seem like overall they're targeting different concerns.

the cwl stuff is targeting similar concerns and is relevant prior art. Exact reproducibility (same inputs, same code versions, same outputs) is the first step in increasing the quality of scientific software. The very next step is ensuring that scientific code is reusable, meaning different inputs, same code, different but correct outputs. Tools like cwl are aimed at increasing the reproducibility and reusability of software tools in scientific computing. Both are relevant and build on each other. I don't think that cwl tools are out of scope for what Dr. Watson should aim to do.

Datseris:
My latest commit addresses @kescobo 's comments, and following @jpfairbanks we will leave CWL in the text.

kescobo:
Let me just clarify my point. And to be clear, it's totally possible I'm missing some aspect of cwl, DrWatson, or both. I totally agree that they both fall under the umbrella of "enabling reproducible science with code," but towards different aspects of that. 

For cwl, the point as I see it is to run the same analysis over and over again. Say, if I'm looking at the stock market, each day I pull data, clean it up in some way, filter it, and aggregate it with some other info about the day (day off week, weather, etc). This is done every day the same way, taking some input streams and generating some outputs.

Dr.Watson comes in after this step, when I'm taking all of the outputs from different days and running some analyses or models on the aggregated output. Sure, I might do this once after collecting 20 days off info, and then want to update it and use all the features to track how it changes when I have 30 days off info, but I don't think I would use DrWatson for the repetitive data-cleaning workflow, and I certainly wouldn't try to write a cwl script for the sort of ad hoc exploratory work that DrWatson seems suited for.

I should also clarify that my point was not to remove cwl from the discussion entirely, but rather to make the point that DrWatson fills a different (and much less occupied IMO) niche. Comparing it directly to cwl detracts from the paper I think, since that functionality is pretty saturated.

But I don't feel super strongly about it - write it how you want 😀

Datseris:
Good points, I had misunderstood. I'll make an adjustment!

Datseris:
3a956aaa25b7ae7b70c8f18f3ba3ae440e120ed4

jpfairbanks:
I guess I see the hierarchy of reproducibility tasks the other way. I would think that you start by “freezing a workflow” into a perfectly reproducible process on the same input and same output. And then move on to making the software more reusable on generic problem similar to the original problem. As this process evolves, a script grows into a library. CWL is an orchestration framework for interoperating a bunch of tools at various stages of maturity. 

I think what separates scientific programming from software engineering is that SE often starts out with the aim of building a reusable library, and scientists start with a very specific implementation for their scientific system and make it more generic over time. In this pattern of development I think the “freeze the analysis” then “make it more reusable” lifecycle is more common than “implementing a generic tool” then “apply the generic tool to a specific problem”. But we might be imagining different  userbases.

tamasgal:
Indeed, the CWL part is a bit whacky, since that’s a framework which is main goal is reproducibility but the method is a different one as it is an abstraction layer between the software (running in a shell or container, basically) and the workflows themselves.
There are definitely similarities regarding the general aim but fundamental differences between the methods.

I’ll try to rephrase it a bit better...

tamasgal:
Done in https://github.com/JuliaDynamics/DrWatson.jl/pull/194

PR #193, opened 2020-09-16
Add coverage

JonasIsensee:


JonasIsensee:
@Datseris 

codecov seems to be running checks but I think you need to authorize the codecov app before it starts posting anything here.

Datseris:
Hi Jonas, I've added CodeCov to the organization, is there something more necessary?

JonasIsensee:
Actually I must admit, I don't really know.
I think this should start working properly once we merge.
(e.g. the badge only works once codecov has run on the master branch which won't happen before we merge this)

PR #196, opened 2020-09-20
Add a "sort" option to `savename`

tshort:
Sometimes, it's nicer to have names sorted in a particular order with more important parameters at the front. This makes it easier when viewing outputs in a file manager, too. 

Obviously, this only works for types like NamedTuple that are sorted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 1.16.0
+* Add a `sort` option to `savename`.
 # 1.15.0
 * Better default readme file for a new project.
 # 1.14.6
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.15.1"
+version = "1.16.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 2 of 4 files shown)


Datseris:
Seems reasonable to me. Can you please increment minor version in Project.toml file and add a mention in CHANGELOG.md?

Datseris:
Ah wait: what does `sort =false` do for dictionary inputs?

tshort:
For a dictionary, the order will be random. That's obviously not very useful. You'd only use it for a type that has ordering, like a NamedTuple or one of the ordered dictionary types in OrderedCollections.jl or Dictionary.jl.

tshort:
With a dictionary, you can also specify ordering with the `accesses` argument.

Datseris:
> random

Hm, that's not good. For me, one of the biggest points of `savename` is that it is guaranteed to be deterministic... So If someone uses `sort=false` with dictionary, that breaks this promise. What is the best way to go about this?

tshort:
The easiest option is to document that the "sort" keyword argument shouldn't be used with unordered collections unless the "accesses" argument is given.

If there was a trait or some other way to determine if a type had deterministic ordering, we might be able to use that, but I don't think that exists.

Datseris:
Okay I think this is a good idea. I'd propose to say in the description of sort:

 Indicate whether the pairs are sorted alphabetically by keys. If not, they are sorted by the order of `accesses`. WARNING: the default `accesses` is not deterministic for `Dict` inputs.

Datseris:
Also, don't use the word "always" when we say that we sort alphabetically

tshort:
@Datseris, I updated this PR based on your requests.

PR #202, opened 2020-10-22
Update sumatra-part

tamasgal:
Targets the Sumatra part in the latest review comment on the paper: https://github.com/openjournals/joss-reviews/issues/2673#issuecomment-714517043

### DIFF
--- a/paper/paper.bib
+++ b/paper/paper.bib
@@ -104,3 +104,18 @@
   title = {{CaosDB}{\textemdash}Research Data Management for Complex,  Changing,  and Automated Research Workflows},
   journal = {Data}
 }
+
+@inproceedings{ChirigatiRSF16,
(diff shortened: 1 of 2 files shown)


Datseris:
Can we add the 

> I think ReproZip and ActivePapers should be mentioned in the "Comparison" section, as tools which also aim to make scientific projects easily portable, but in more language-independent ways.

part here almost verbatim? Seems like the simplest way to address.

Datseris:
Also, _faaaaast_.

tamasgal:
Not sure about ActivePapers; that project seems quite dead to me...

Datseris:
> Not sure about ActivePapers; that project seems quite dead to me...

That is good enough justification. After all we have to _reply_ to a review, not absolutely do what it says.

tamasgal:
:+1:

Issue #204, opened 2020-10-23
dict_list order changes on different machines


lhupe:
The output order of `dict_list` is different on different devices.

## MWE
```julia
using DrWatson

d = Dict(
    :foo => [0.01, 0.05],
    :bar => [0.1, @onlyif(:foo == 0.01, 0.4)],
    :baz => Inf
)
simgrid = dict_list(d)

println.(simgrid)
```

produces 

```
Dict(:baz => Inf,:bar => 0.1,:foo => 0.05)
Dict(:baz => Inf,:bar => 0.1,:foo => 0.01)
Dict(:baz => Inf,:bar => 0.4,:foo => 0.01)
```
on one machine and
```
Dict(:baz => Inf,:bar => 0.1,:foo => 0.01)
Dict(:baz => Inf,:bar => 0.1,:foo => 0.05)
Dict(:baz => Inf,:bar => 0.4,:foo => 0.01)
``` 
on another. 

Both machines use the official Julia v1.5.2 binaries, but have different operating systems and hardware configurarions.

This appears to be an extremely arcane bug — e.g. I get identical results on both machines by changing `baz => Inf` to e.g. a finite float. I wouldn't be surprised if it's hard to reproduce.

tamasgal:
Julia's `Dict` is not ordered, so this is definitely not an "arcane bug" 😉 I think it's more worrying that you rely on the order of something which does not give you a guarantee on that by design.

Let me ask differently: why do you need the exact same order?

There is also a Dict-like data structure  which gives you guaranteed order, called `OrderedDict` and available in https://github.com/JuliaCollections/DataStructures.jl

tamasgal:
EDIT (deleted): sorry I misread -- long night of coding 😅

Datseris:
Regardless, I am not sure whether this is a "bug"...? Also Lukas: why is it important for you? What's the application?

tamasgal:
So in principle it does not count as a "bug" as there is no mentioning of any order (perseverance) in the docs: https://juliadynamics.github.io/DrWatson.jl/dev/run&list/#DrWatson.dict_list

lhupe:
1) as far as I can tell (with `dump`), the representation of the dictionaries is identical for both machines.
2) yes, it's not an actual "bug" by definiton, but it would be nice if the output of the function were stable. I discovered this when playing around with some of my data, and I got quite a shock when the parameters stored in the data file did not match my expectations from the `dict_list`.

[...]

PR #205, opened 2020-10-23
replace Set with unique!

lhupe:
see #204

Tests passed locally


Datseris:
please increment patch version number in Project.toml

PR #210, opened 2020-11-09
Fix treatment of nested onlyif definitions

sebastianpech:


sebastianpech:
@JonasIsensee Turns out it was easier than I thought (unless I missed something). My initial assumption was that this can't be done because `@onlyif` can't know if it is inside a vector or not. This is true, however, there is a stage where I see this vector containing the `@onlyif`, which is right before generating the list. So the fix is basically to check if there are any `Vector{Vector{DependentParameter}}` (which is not allowed) and if there are undo the expansion that `@onlyif` did when it was first called.

sebastianpech:
@Datseris do you know what die Travis CI - Branch thing does? I accidentally pushed my branch to the main repo after I opened the PR from my fork (I know I could always use the main repo, but I still prefer the fork ;)) and then deleted the branch. Now that build is failing ...

Datseris:
No clue sorry :( But I know that we have to transition away from Travis sooner or later, see https://github.com/JuliaDynamics/Agents.jl/issues/340   :(

JonasIsensee:
Hi @sebastianpech ,

thanks for implementing this!
I read through the changes and it looks good.

I also tested locally and it does what it is meant to but I must admit that for me personally it is not very intuitive. (I'm aware and sorry that this was one of the solutions that I suggested)

```julia
julia> dict_list(Dict(:a=>[1,2], :b=>@onlyif(:a==1, [3,4])))
3-element Array{Dict{Symbol,Int64},1}:
 Dict(:a => 1,:b => 3)
 Dict(:a => 2)
 Dict(:a => 1,:b => 4)

julia> dict_list(Dict(:a=>[1,2], :b=>[@onlyif(:a==1, [3,4]), @onlyif(:a==2, [5,6])]))
2-element Array{Dict{Symbol,V} where V,1}:
 Dict{Symbol,Any}(:a => 1,:b => [3, 4])
 Dict{Symbol,Any}(:a => 2,:b => [5, 6])

```

This led me to test whether my idea was feasible and I ended up with this
similarly simple change:

```julia
julia> dict_list(Dict(
    :a=>[1,2], 
    :c => [ 5, @onlyif( :a == 1, [3,4] ), @onlyif( :a == 2, [[6,7]] ) ]
))
5-element Array{Dict{Symbol,V} where V,1}:
 Dict(:a => 1,:c => 5)
 Dict(:a => 2,:c => 5)
 Dict(:a => 1,:c => 3)
 Dict(:a => 1,:c => 4)
 Dict{Symbol,Any}(:a => 2,:c => [6, 7])
```
which is reached by implementing and running this on `iterable_dict` in `_dict_list`.
```
function resolve_nested_conditions!(c::Dict)
    modified = false
    for key in keys(c)
        c[key] = mapreduce(vcat, c[key]) do entry
            if entry isa Vector{<:DependentParameter}
                entry
            else
                [entry]
            end
        end
    end
    return c
end
```

Your PR is ready for merge if that is the behaviour that we find most useful.
Please let me know, what you think!

Jonas

sebastianpech:
Thanks @JonasIsensee for reviewing. I thought about that case you're describing too but decided against implementing this (though I'm not 100% sure it is the best decision). 
I'm thinking about how `@onlyif` acts always in the context of `dict_list` which led me to the decision that **`@onlyif` should only expand if it acts on a vector that would also be expanded by `dict_list`.**
However, as we know that `@onlyif` actually always expands, you can use splatting to do what you suggested in your example:
```julia
julia> dict_list(Dict(
    :a=>[1,2], 
    :c => [ 5, @onlyif(:a == 1, [3,4])..., @onlyif( :a == 2, [6,7] ) ]
))

5-element Array{Dict{Symbol,V} where V,1}:
 Dict(:a => 1,:c => 5)
 Dict(:a => 2,:c => 5)
 Dict(:a => 1,:c => 3)
 Dict(:a => 1,:c => 4)
 Dict{Symbol,Any}(:a => 2,:c => [6, 7])
```
What do you think about that?

JonasIsensee:
Hi @sebastianpech ,

this is neat! I didn't know you could just splat the `@onlyif` stuff. 

Your explanation makes sense - count me convinced. :)

JonasIsensee (approved):

JonasIsensee:
Also, I'm playing around with this again...

I found that if we just add one single definition
```julia
function DependentParameter(value::DependentParameter, condition::Function)
    new_condition = (args...) -> (condition(args...) && value.condition(args...))
    DependentParameter(value.value, new_condition)
end
```
suddenly we can do nested `@onlyif`s.

```julia
julia> dict_list(Dict(
    :a=>[1,2], 
    :b => [3,4], 
    :c => @onlyif( :a == 2, [5, @onlyif(:b == 4, 6)])
))
5-element Array{Dict{Symbol,Int64},1}:
 Dict(:a => 1,:b => 3)
 Dict(:a => 2,:b => 3,:c => 5)
 Dict(:a => 1,:b => 4)
 Dict(:a => 2,:b => 4,:c => 5)
 Dict(:a => 2,:b => 4,:c => 6)
```

sebastianpech:
Ok, that's brilliant. Can't think of any documented use cases where this might lead to a problem. I think we should definitely add it.

JonasIsensee:
Awesome.
This is your PR opened from your fork. I'm not sure I can push that change in.
Do you want to just add that definition and tag me in the commit message?

The example I gave is probably a decent test.

sebastianpech:
I will tag you in the commit msg

PR #217, opened 2020-11-24
added an option to parse parameters as Named Tuple

pevnak:
This is a tiny PR, which adds an option to return parsed parameters in `parse_savename` as NamedTuple. It of course causes type instability, but I do not think it will be devastating in this case.

### DIFF
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -359,10 +359,12 @@ is delimited by `=` and the closest `connector`. This allows the user to have `c
 * `connector = "_"` : string used to connect the various entries.
 * `parsetypes = (Int, Float64)` : tuple used to define the types which should
   be tried when parsing the values given in `filename`. Fallback is `String`.
+* `asnamedtuple = false` : return parsed arguments as `NamedTuple` instead of `Dict`
 """
 function parse_savename(filename::AbstractString;
                         parsetypes = (Int, Float64),
-                        connector::AbstractString = "_")
+                        connector::AbstractString = "_",
+                        asnamedtuple = false)
     length(connector) == 1 || error(
     "Cannot parse savenames where the 'connector'"*
     " string consists of more than one character.")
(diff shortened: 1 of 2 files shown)


Datseris:
Thanks for opening a PR!

What is not clear to me is why would this be part of the API, instead of being done by the user after they have obtained the output of `parse_savename`. What's the justification? If I understood correctly, changing the output is essentially a one-liner:

`parameters = asnamedtuple ? (;map(x -> (Symbol(x[1]), x[2]), collect(parameters))...) : parameters`

pevnak:
Thanks for an attention. It is a matter of convenience. I tend to (over)use NamedTuples and whenever I would be parsing, I would need to add this line. Thus having this as a part of API is a matter of convenience. Even though this is a one liner, it removes the visual clutter in scripts, where this is used.

Datseris:
> Even though this is a one liner, it removes the visual clutter in scripts, where this is used.

You could define your own function however that does this, and then use this function everywhere without increasing clutter.

It's just that what this PR does is weird from an API perspective. It gives an option to fundamentally change the return type of a function, which for the Julia language is very atypical. Sure, for the current function this type instability doesn't matter, as the function is doing parsing work and is thus type unstable de facto, but it still seems weird for me.

I've asked a second review by @sebastianpech .

sebastianpech:
I get that this would be handy to have, but from the API perspective, I don't think it is a good idea to over-specialize things. However, what we could do is change the Dict type to return with symbol keys. (I don't remember why we went with strings?) This way your code would actually simplify to

```
julia> d = Dict(:a=>10, :b=>11)
Dict{Symbol,Int64} with 2 entries:
  :a => 10
  :b => 11

julia> (;d...)
(a = 10, b = 11)
```

Though this might be a breaking change, so maybe not such a good idea after all.
@pevnak Have you looked at defining default functions for your project: https://juliadynamics.github.io/DrWatson.jl/dev/real_world/#Making-your-project-a-usable-module-1

pevnak:
Thanks for pointing me to this, I will investigate it. Meanwhile, I will close this.

Datseris:
@pevnak I'm really sorry that your first PR in this project end up being closed. Please don't take this at heart, and I hope that in the future if you have ideas you can consider contributing them here. If you feel like you don't want to "risk" spending time, feel free to open an issue to simply propose your idea.

PR #218, opened 2020-11-27
move to github actions

Datseris:

PR #221, opened 2021-01-09
Two dict key type exchange functions

Datseris:
I find them convenient, especially when working with JLD2, and they kinda make sense given that we have `@dict, @strdict, ntuple2dict`.


Datseris:
Requires #218 first.

PR #225, opened 2021-01-29
Fix #223 and #224

### LINKED ISSUE #223: dict_list_count returns wrong number of elements with @onlyif
`dict_list_count` shows a different number than the resulting size of `dict_list`.

Here is a MWE : 

```julia
dict = Dict(
    :a => [1, 2, 3],
    :b => [1:9..., @onlyif(:a > 1, collect(10:10:99))..., @onlyif(:a > 2, collect(100:100:1000))...] # I did not find a nicer way to write this
)

length(dict_list(dict)) == 55
dict_list_count(dict) == 84
```

With julia 1.5 and DrWatson 1.16.6

### LINKED ISSUE #224: Submodules flag not working in older versions of git.
**Describe the bug**
When using `git version 1.8.3.1`, `gitpatch` causes
```
fatal: Failed to parse --submodule option parameter: 'diff'
┌ Warning: `gitpatch` failed with error ProcessFailedException(Base.Process[Process(`git --git-dir=<my_path>/.git --work-tree=<my_path> diff --submodule=diff HEAD`, ProcessExited(128))]), returning `nothing` instead.
```

**Minimal Working Example**
```julia
using DrWatson
DrWatson.gitpatch()
```

**DrWatson.jl version**
DrWatson, version 1.16.5 and newer.
Caused by https://github.com/JuliaDynamics/DrWatson.jl/pull/212
Git version 1.8.3.1.
Works on 1.16.4.

Would it be possible to check for the git version and use newer version with submodules only when git itself supports it, so it wouldn't break when using older git?
Thanks.

sebastianpech:
Closes #223, Closes #224

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.16.7
+* Fix `dict_list_count` returns wrong number of elements with `@onlyif`. (#223)
+* Fix `gitpatch` for git versions not supporting submodules. (#224)
 # 1.16.6
 * `@onlyif` now doesn't expand vector arguments if it's placed inside another vector. (#209)
 * `@onlyif` now supports chaining. See real world examples: "Defining parameter sets with restrictions". (#210)
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.16.6"
+version = "1.16.7"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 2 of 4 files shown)


sebastianpech:
I checking the message of stderr from the git process contains both --submodule and diff (To circumvent problems with error messages in other languages and also not needing to find out which version of git supports submodule ;) ). If they are both contained, the function now retries without the submodule option. If it still fails the error is printed as before.

sebastianpech:
@racinmat can you maybe check if that works on your machine?

Datseris (approved):
Gotta admit, this CodeCov stuff is making reviews much harder than they should be.

Datseris:
@sebastianpech what's the status here? is this good to go? (If someone approves the PR and you think it is good to go, feel free to merge at your leisure!)

sebastianpech:
@Datseris I was hoping that maybe @racinmat might verify that the code works on the older git version. Will merge in the next days.

racinmat:
I'll verify it on monday.

racinmat:
So, I'm confirming that with this fix, `gitpatch()` works, while on the newest stable DrWatson version it does not. 
Looking forward to see this merged.

Issue #126, opened 2020-02-09
Argument order for `produce_or_load`


xukai92:
It might be nice to move the function argument `f` as the first agument of `produce_or_load` so people can use the `do` keyword instead of defining the function explicitly

```julia
N = 2000; T = 2000.0
file = produce_or_load(
    datadir("mushrooms", "toy"), # path
    @dict(N, T), # container
    prefix = "fig5_toyparams" # prefix for savename
) do d
    HTEST = 0.1:0.1:2.0
    WS = [0.5, 1.0, 1.5]
    @unpack N, T = d
    toypar_h = [[] for l in HS]

    for (wi, w) in enumerate(WS)
        println("w = $w")
        for h in HTEST
            toyp = toyparameters(h, w, N, T)
            push!(toypar_h[wi], toyp)
        end
    end
    return @dict toypar_h
end
@unpack toypar_h = file
```

asinghvi17:
Looks like it would be as simple as defining:
```julia
produce_or_load(f::Function, c) = produce_or_load(c, f)
produce_or_load(f::Function, p::String, c) = produce_or_load(p, c, f)
```
?

Datseris:
Yes seems to be that simple. @xukai92 if you test this extention and it works, we should put it in the docs! Seems to be a nice convenient syntax!

xukai92:
It actually works nicely with an extra `; kwargs...` in all function calls.

> Unfortunately changing the order of arguments without this dispatch example is a breaking change ...

What does this mean?

Datseris:
You (meaning "we", the developers) cannot change the order of the arguments of a function used by users. It will break their code. Anyone that uses `produce_or_load` will have a code that doesn't work if we just change the order of arguments around.

xukai92:
Oh I see. You mean **without** the dispatch. Is it OK to just add the dispatch version then? Seem to solve the problem by itself.

Datseris:
I think adding the dispatch is _probably_ fine. It will create ambiguity errors though when users pass `c <: Function` into `produce_or_load`. It is possible to create custom structs and subtype them as functions.

If you want to open a PR about this, we should test what happens in this scenario where a user tries to use a `c <: Function`.

xukai92:
What's the status of 2.0? Is this too late to add this change (or it's already there)?

Datseris:
[...]

PR #227, opened 2021-02-02
Support do block for `produce_or_load`

xukai92:
Addressing https://github.com/JuliaDynamics/DrWatson.jl/issues/126

### DIFF
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -27,7 +27,9 @@ the macros [`@dict`](@ref) and [`@strdict`](@ref) can help with that.
 See also [`savename`](@ref).
 """
 produce_or_load(c, f; kwargs...) = produce_or_load("", c, f; kwargs...)
-function produce_or_load(path::String, c, f;
+produce_or_load(f::Function, c) = produce_or_load(c, f)
+produce_or_load(f::Function, path::String, c) = produce_or_load(path, c, f)
+function produce_or_load(path::String, c, f::Function;
     tag::Bool = true, gitpath = projectdir(), loadfile = true,
     suffix = "bson", prefix = default_prefix(c),
     force = false, verbose = true, kwargs...)
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -11,6 +11,12 @@ that produces your data. Then save `file` as `s` and then return `file, s`.
 The function `f` must return a dictionary,
 the macros [`@dict`](@ref) and [`@strdict`](@ref) can help with that.
 
+You can use [do-block](https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments) instead of defining a function to pass in. For example,
+```julia
+produce_or_load(c) do c
+    # simulation wiht config `c` runs here
+end
+
 ## Keywords
 * `tag = true` : Save the file using [`tagsave`](@ref).
 * `gitpath = projectdir()` : Path to search for a Git repo.
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -13,7 +13,7 @@ the macros [`@dict`](@ref) and [`@strdict`](@ref) can help with that.
 
 You can use [do-block](https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments) instead of defining a function to pass in. For example,
 ```julia
-produce_or_load(c) do c
+produce_or_load([path="",] c) do c
     # simulation wiht config `c` runs here
 end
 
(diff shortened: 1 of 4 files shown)


xukai92:
I guess it would be interesting to add a tip in the doc as well. How about something like below?

---
## Tips

You can use [do-block](https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments) instead of defining a function to pass in. For example,
```julia
produce_or_load(simulation) do simulation
    # simulation happens here
end
```
---

Datseris:
Yeap, I would suggest the same, add information about this in the documentation string. You don't even have to create a `Tips` section, just mentioned it before the keywords start..

xukai92:
Added!

src/saving_files.jl
```diff
@@ -11,6 +11,12 @@ that produces your data. Then save `file` as `s` and then return `file, s`.
 The function `f` must return a dictionary,
 the macros [`@dict`](@ref) and [`@strdict`](@ref) can help with that.
 
+You can use [do-block](https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments) instead of defining a function to pass in. For example,
+```julia
+produce_or_load(c) do c
+    # simulation wiht config `c` runs here
+end
```
Datseris, line 18: What if I do `produce_or_load(path, c)` ? What comes after the `do`?

xukai92: Clarified!

Datseris: But are you sure that it is `c` that comes after the do if I use a 2-argument version?

test/savefiles_tests.jl
```diff
@@ -78,6 +78,24 @@ for ending ∈ ("bson", "jld2")
...
+    @test !isfile(savename(simulation, ending))
+    sim, path = produce_or_load(simulation; suffix = ending) do simulation
+        @test typeof(simulation.T) <: Real
+        a = rand(10); b = [rand(10) for _ in 1:10]
+        return @strdict a b simulation
+    end
+    @test isfile(savename(simulation, ending))
+    @test sim["simulation"].T == T
+    @test path == savename(simulation, ending)
+    sim, path = produce_or_load(simulation; suffix = ending) do simulation
```
Datseris, line 90: for safety here you should use the `produce_or_load(path, simulation; ...) do ...` syntax, so that both are checked.

xukai92: Should we do the same for the exisiting tests that don't use do-block as well then?

Datseris: I don't think that is necessary. Here it is simply not clear to me of `f(a, b) do x` what is `x`? is it the tuple `(a, b)` or is it only `a`? Or is it `b`?

xukai92: OK. I updated the test. Although it should not be an issue to people familar with do-blocks I suppose because they themselves have nothing to do with other arguments in the function call.

Datseris: You have written: `f(path, c) do c`, which confuses me, because `c` is the same symbol as the input argument. How does this work? What is `c`? The Julia docs provide the example:

```
open("outfile", "w") do io
    write(io, data)
end
```
here `io` is neither `"outfile"` nor `"w"`, so how in our case does what comes _after_ the `do` block coincide with the _input argument_ given to the function that initializes the `do` block?

src/saving_files.jl
```diff
@@ -11,6 +11,12 @@ that produces your data. Then save `file` as `s` and then return `file, s`.
 The function `f` must return a dictionary,
 the macros [`@dict`](@ref) and [`@strdict`](@ref) can help with that.
 
+You can use [do-block](https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments) instead of defining a function to pass in. For example,
+```julia
+produce_or_load([path="",] c) do c
+    # simulation wiht config `c` runs here
+end
+
```
Datseris, line 19:
```suggestion
```
```

Datseris, line 19: here we miss three `

xukai92: Thanks for finding this! Added.

Datseris:
Okay I finally understand what `do` blocks do with `c`. It is fine as-is now but you have to increment minor version in Project.toml and also add an entry in the changelog.md.

Other than that, no other comments, thanks a lot for the contribution!

xukai92:
Thanks for the review! Just incremented minor version and updated the changelog.

Datseris:
Sorry for the confusion, the version needs to become 1.17.0, the minor number is the one in the middle, and is incremented when a backwards compatible new feature is added.

xukai92:
Fixed. Thanks for the explanaiton - I should have known this!

Issue #232, opened 2021-03-01
Would possible to use engineering notation for floating-point numbers?


PetrKryslUCSD:
**Is your feature request related to a problem? Please describe.**
Large or small numbers make for bad simulation names. This might help.

**Describe the solution you'd like**
Example: `sim_E=2.1e11`.

**Describe alternatives you've considered**

sebastianpech:
I'm closing this in favor of the discussion in the referenced issue.

PR #233, opened 2021-03-01
Rounding and scientific notation in `savename`

### LINKED ISSUE #158: savename improvement for scientific notation
If it uses `scientific`, does it convert integers like `10^10` to scientific notation?

### LINKED ISSUE #163: treatment of floating points in savename
When the value of a parameter is a float, `savename` checks if it can be rounded to an integer. This is good for keeping the filename short. But at the same time, it makes it cumbersome to work with the filename outside of DrWatson (e.g. checking whether the file exists in shell, etc.). I wonder if it may be a good idea to add an optional argument that controls the number of digits for floating points.

Datseris:
Closes #163, closes #158

Okay, the hope here is to address all concerns and requests of #163. It also removes the "Float -> Integer" existing convention of `savename`, which in hindsight was a bad idea.

Currently work in progress.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 1.19.0
+* `savename` will no longer replace a floating point number with its integer version, if they coincide with respect to rounding. Thus, "integer" `AbstractFloat`s will always end with `.0` in `savename`.
 # 1.18.3
 * Remove type constraints on `produce_or_load` path argument (#229)
 # 1.18.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 * `tmpsave` now saves as JLD2.jl and requires loading explicitly the `"params"` field.
 * `collect_results` saves as JLD2.jl by default now.
 * `produce_or_load` saves as JLD2.jl by default now.
-
-# 1.19.0
 * `savename` will no longer replace a floating point number with its integer version, if they coincide with respect to rounding. Thus, "integer" `AbstractFloat`s will always end with `.0` in `savename`.
+* `savename`'s `scientific` keyword has been replaced by `sigdigits`, as this was the only thing it was doing.
+
 # 1.18.3
 * Remove type constraints on `produce_or_load` path argument (#229)
 # 1.18.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 * `produce_or_load` saves as JLD2.jl by default now.
 * `savename` will no longer replace a floating point number with its integer version, if they coincide with respect to rounding. Thus, "integer" `AbstractFloat`s will always end with `.0` in `savename`.
 * `savename`'s `scientific` keyword has been replaced by `sigdigits`, as this was the only thing it was doing.
+* `savename`'s default options now have `sigdigits = 3` instead of `digits = 3`.
+## Non-breaking
+* `savename` now allows a keyword `val_to_string` for customization.
 
 # 1.18.3
 * Remove type constraints on `produce_or_load` path argument (#229)
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.18.3"
+version = "1.19.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/DrWatson.jl
+++ b/src/DrWatson.jl
@@ -28,7 +28,7 @@ _wsave(filename, data...; kwargs...) = FileIO.save(filename, data...; kwargs...)
 
 Save `data` at `filename` by first creating the appropriate paths.
 Default fallback is `FileIO.save`. Extend `wsave` for your type
-by extending `DrWatson._wsave`.
+by extending `DrWatson._wsave(filename, data...; kwargs...)`.
 """
 function wsave(filename, data...; kwargs...)
     mkpath(dirname(filename))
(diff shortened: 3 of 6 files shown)


Datseris:
So, how do we do the original suggestion of #163 from an API perspective?

sebastianpech:
As I wrote in the original suggestion, in my opinion, it is a breaking change thinking of `produce_or_load`. Besides redoing simulations, a user could solve this by parsing the original savename, loading the data and saving with the new `savename` function.

sebastianpech:
What `scientific` does is really misleading, I'm wondering why we named it that. It is just passed to `sigdigits` in `round` and does nothing else. You get scientific notation also without it:

```
julia> savename("sim",(b=0.000000001,),digits=10)
"sim_b=1e-9"
```

This means, removing the integer actually influences all uses, not just the ones using `scientifc`.
This also means: A better name for `scientific` is `sigdigits`.

sebastianpech:
> This will most likely necessitate a 2.0 or the explicit use of an extra keyword (which I don't like, savename is already quite heavy).

I also don't like adding a new keyword with some strange switch that activates integer printing. As much as I would like to simply change those few lines of code, I think a clean implementation breaks too much.

Datseris:
Okay, let's go for 2.0 then. Let's gather any other breaking changes that come to mind (if any, I don't think so). Is `sigdigits` necessary? What do we do to satisfy what users in #163 and #232 asked for...?

Datseris:
I personally did not find anything else worthy of 2.0. To recap, we also need to close https://github.com/JuliaDynamics/DrWatson.jl/issues/158 with this one.

sebastianpech:
#158 **should be fine**
If the number is of type Int, scientific notation is not used.
```
3*10^7 -> 30000000
3.0*10^7 -> 3.0e7
```

#233 **should be fine**
For large and for small float numbers scientific notation is used.

#163 **partly solved**
I guess a solution that solves this issue would mean adding an additional argument for how float numbers are formatted, so you would be able to print something like `1.00`. Maybe that's not such a bad idea. This way users could easily implement the old behavior and not break their code, by passing a function that does the integer conversion thing.

Datseris:
Do we even know what formatting function always prints number with a specific number of `0`s...? I know the *macro* `@printf`, but you can't be passing macros around.

Datseris:
Okay, so #234 is also breaking. Let's merge them together and be done with it. Problem is, I'm really not sure what remains to be done here. I will rename `scientific` to `sigdigits`, but other than that, I really don't know.

sebastianpech:
Yeah I think it's good this way. Let's not overcomplicate savename.

George Datseris <notifications@github.com> schrieb am So., 7. März 2021,
18:59:

Datseris:
Something that is not currently happening: printing the output with given "digits" no matter what `round` does, e.g. what the user requested to print `1.0` as `1.00` etc. 

Should we go into any length doing this? Or should we just say that `savename` prints whatever `random` gives?

Datseris:
@tamasgal , @JonasIsensee as this is the second breaking change for 2.0, would be good to get your input. Please have a quick look over this PR.

So, at the moment, there have been various requests to further tune how `savename` outputs floats, e.g. to "somehow" make it *always* use scientific notation a-la `3e-1`, even when Julia doesn't use it by default, or to allow it to somehow output numbers with always a specific amount of characters (e.g. always do `0.01`, `0.00`, `0.10`, etc. for 3 characters. 

What do you think of these things? I'm not sure whether I'd like to support these to be honest, and I'm also not sure whether we should further complicate `savename`, which already has a lot of arguments. A user can in principle implement their own number type and extend `string` to be able to do whatever they want for printing, but that's also probably going too far.

JonasIsensee:
> @tamasgal , @JonasIsensee as this is the second breaking change for 2.0, would be good to get your input. Please have a quick look over this PR.
> 
> So, at the moment, there have been various requests to further tune how `savename` outputs floats, e.g. to "somehow" make it _always_ use scientific notation a-la `3e-1`, even when Julia doesn't use it by default, or to allow it to somehow output numbers with always a specific amount of characters (e.g. always do `0.01`, `0.00`, `0.10`, etc. for 3 characters.
> 
> What do you think of these things? I'm not sure whether I'd like to support these to be honest, and I'm also not sure whether we should further complicate `savename`, which already has a lot of arguments. A user can in principle implement their own number type and extend `string` to be able to do whatever they want for printing, but that's also probably going too far.

Hi @Datseris ,
I think that these changes are quite very reasonable.
I would also think that the two additional features you mention, would be very nice to have.
In particular for viewing in a terminal or whatever it would be nice to have a very predictable 
number of characters across different ranges of parameter values.

However, I've never actually used `savename` much (either computations take too long or have too many parameters),
so I don't know whether the added complexity would be worth it. Honestly, I'm not even sure whether such a general
solution exists.

sebastianpech:
We could do something kind of like the [typemap in CSV.jl](https://csv.juliadata.org/stable/#Typemap). Instead of a target type on passes their own implementation of `roundval` or a string readable by `@printf`. Besides that, I think there are no other good solutions. One alternative would be to allow custom definitions just like we did with `_wsave`. For this, however, we would need to run every value though `roundval`.

Datseris:
> We could do something kind of like the typemap in CSV.jl. 

I think that is way too complicated for a function that is supposed to be a simple shorthand and for convenience.

Here is what I propose: allow users to provide a function that transforms any value into a string. By default this will be what we have now: if float attempt to round, otherwise call `string(val)`.

> One alternative would be to allow custom definitions just like we did with _wsave. 

In essense my proposal does the same thing. Furthermore, the users can already overload `Base.string` to do the same.

Datseris:
The last commit implements my proposal

JonasIsensee:
> In essense my proposal does the same thing. Furthermore, the users can already overload `Base.string` to do the same.

Hmm, that might propose a very strong case of type piracy. I don't think that is a good idea.

PR #234, opened 2021-03-05
replace bson with jld2

### LINKED ISSUE #231: Switch all doc to use JLD2.jl instead of BSON.jl
Using BSON.jl in our docs leads newcomers of Julia to using BSON instead of JLD2, which I think is really not a good idea... We should really change it.

p.s.: I'm really surprised that new users come to DrWatson first before even checking out a package to save files :D

Datseris:
Closes #231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 2.0
+## Breaking
+* DrWatson has moved entirely on using JLD2.jl instead of BSON.jl for saving files, and also suggests the same to its users through the documentation.
+* `tmpsave` now saves as JLD2.jl and requires loading explicitly the `"params"` field.
+* `collect_results` saves as JLD2.jl by default now.
+* `produce_or_load` saves as JLD2.jl by default now.
+
 # 1.18.3
 * Remove type constraints on `produce_or_load` path argument (#229)
 # 1.18.2
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "1.18.3"
+version = "2.0.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,11 +1,11 @@
 [deps]
-BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterTools = "35a29f4d-8980-5a13-9543-d66fff28ecb8"
 DrWatson = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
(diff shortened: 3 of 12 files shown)


Datseris:
@JonasIsensee could you please help here? I dont remember fully; does JLD2 require that dictionaries must have String type as keys?

JonasIsensee:
Hi @Datseris ,

yes, JLD2 can only store string keys and it is not possible to change that internally
since it follows the HDF5 standard which has an internal file system like representation.


If you want to use `Symbol`s in DrWatson, we have a few workaround options. 
1) Write a new API function that automatically converts to string when writing. The problem is that you need to be explicit about this when loading.

2) Write a new API function that prepends some signal characters e.g. `:param` → `"__param"` which can then be parsed back to symbols upon loading. 
Potential problem:
- If implemented in DrWatson, then the data is different when loaded directly.
- If implemented in JLD2, then it still doesn't integrate so well with the path like structure.

3) JLD2 can store most objects - just not as keys. Something like 
```
function wsave(filename, data...; kwargs...)
   if endswith(filename, ".jld2")
        d = Dict(data)
        FileIO.save(filename, "data", d; kwargs....)
    end
end
```

JonasIsensee:
Step three has the disadvantage that it reduces performance for large files as it becomes impossible to load only parts of a file.
Step two breaks the type stability for the keys of loaded files but maybe that would be acceptable.

JonasIsensee:
I'm implementing something like idea no 2 in https://github.com/JuliaIO/JLD2.jl/pull/287

I don't yet know whether this is really a good idea, though.

Datseris:
wait, which part of DrWatson doesn't allow us to use string keys? I thought it was a simple change from my end, only in the example files. Am I missing something?

JonasIsensee:
I'm sorry if that wasn't clear.
there is a problem on the JLD2 side, IF you want to use `Symbol` keys as you currently do with BSON.

Datseris:
okay. let me first make this PR working (the examples do not work at the moment), and then we can see what is necessary in JuliaIO/JLD2.jl#287 and whether it is necessary at all

Datseris:
@JonasIsensee I've fixed the `workflow.jl` file in the docs, and now it does correctly use `JLD2`. However, it seems that `collect_results!` is not working. Does it make any assumptions regarding the key type?

Datseris:
Okay @JonasIsensee everything is fine. I was doing something else wrong. I'll now update the tests to also use JLD2 instead of BSON and we are good to go. I don't think the modification you are doing in JLD2.jl is necessary and as you said it also has some problems, so I think you can close that PR.

JonasIsensee:
> Okay @JonasIsensee everything is fine. I was doing something else wrong. I'll now update the tests to also use JLD2 instead of BSON and we are good to go. I don't think the modification you are doing in JLD2.jl is necessary and as you said it also has some problems, so I think you can close that PR.

Alright, awesome.
I fixed two more things.
`produce_or_load` still had `bson` as the default suffix and
I updated the real world example. 

That example also shows the string-key limitation of jld2. (That isn't really a big problem)

```
return @dict toypar_h
```

has to now say 
```
return @strdict toypar_h
```

but since 
```
@unpack toypar_h = data
```
doesn't care whether `data` contains string or symbol keys, it's not really important.

JonasIsensee:
hm, `tmpsave` still uses `bson` as the default. 
That will break since `BSON` is no longer a dependency of DrWatson

Datseris:
then we merge it with the `savename` PR for 2.0.

JonasIsensee:
Alright.

I implemented my suggestion in the latest commit
and adapted test + real_world example

Datseris:
you need to update the docstring of `tmpsave` accordinhgly though

PR #236, opened 2021-03-17
Fix loading of results collection with JLD2

dominformant:
`collect_results!` is not working with JLD2 files, as the resulting DataFrame is saved in a Dictionary with Symbol keys. Changing this to a String key should have no impact to users, as the base Dictionary is not returned at all.

### DIFF
--- a/test/update_results_tests.jl
+++ b/test/update_results_tests.jl
@@ -102,7 +102,7 @@ end
 #                           Load and analyze  DataFrame                       #
 ###############################################################################
 
-df = BSON.load(defaultname)[:df]
+df = BSON.load(defaultname)["df"]
 @test size(df) == size(cres2)
 @test sort(names(df)) == sort(names(cres2))
 
--- a/test/update_results_tests.jl
+++ b/test/update_results_tests.jl
@@ -8,153 +8,148 @@ using BSON, DataFrames, FileIO, JLD2
 #                        Setup Folder structure                               #
 ###############################################################################
 # %%
-    cd(@__DIR__)
-    isdir("testdir") && rm("testdir", recursive=true)
-    mkdir("testdir")
-    initialize_project("testdir"; git = false)
-    quickactivate("testdir")
+cd(@__DIR__)
+isdir("testdir") && rm("testdir", recursive=true)
+mkdir("testdir")
+initialize_project("testdir"; git = false)
+quickactivate("testdir")
 
 ###############################################################################
 #                           Create Dummy Data                                 #
 ###############################################################################
-    mkdir(datadir("results"))
-    cd(datadir("results"))
+mkdir(datadir("results"))
+cd(datadir("results"))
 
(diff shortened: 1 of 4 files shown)


JonasIsensee:
Hi @dominformant ,
indeed, this must have been overlooked in the 2.0 release and should probably be a considered a bugfix.

test/update_results_tests.jl
```diff
@@ -102,7 +102,7 @@ end
 #                           Load and analyze  DataFrame                       #
 ###############################################################################
 
-df = BSON.load(defaultname)[:df]
+df = BSON.load(defaultname)["df"]
```
Datseris, line 105: wait, why is BSON still used here, I thought we changed everything to not use BSON anymore...? I guess we missed this part. Can you please replace all BSON stuff here with JLD2? Also increase patch version in the Project.toml file?

JonasIsensee: BSON is still a test dependency.
We only changed everything to use JLD2 _by default_.
It seems, we should have added tests, though.

Datseris: Hahah that was quite sloppy i guess. Let's use this PR as an opportunity to _replace_ BSON test dependency by JLD2

JonasIsensee: Fair enough.

Since this is your PR @dominformant, if you're keen to get involved you could help us by replacing every usage of BSON by JLD2 in the test suite. If not, I'll probably get around to it next week since this is my fault in the first place.

dominformant: I just noticed, that [savefiles_tests.jl](https://github.com/JuliaDynamics/DrWatson.jl/blob/946400b27343afdc85c5604532ba64ed8b41f874/test/savefiles_tests.jl#L70) loops over the different file types. Shall I also remove this loop, or shouldn't we consider testing for both, `JLD2` and `BSON` as (as far as I understood) `BSON` is still supported by DrWatson, just not as default?

Datseris: Yeah you have a good point. Sorry, my memory regarding these tests is kind of vague, they were written literally years ago. Please do go forward with your proposal which indeed tests all packages, as it was initially supposed to do anyway.

dominformant: I did a first try to test for both file types, but I'm not totally happy with the result, so I will probably try to make it look nicer the next days. However, I wanted to give you the chance to take a look at the changes so that you can correct me if something is absolutely wrong.

Datseris: Looks great to me to be honest

dominformant: Okay, I added some smaller optical corrections, in my opinion this should be ready now, if you have any hints or ideas, just let me know.

JonasIsensee:
This is looking very good!
Is this ready to merge from your side?

dominformant:
If the question was to me, yes, in my opinion this should be ready to be merged.

dominformant:
Actually now it is just ready now :-D

Issue #239, opened 2021-03-25
Macro Hygiene @strdict


JonasIsensee:
Hi!

I just noticed the following:

```
julia> using DrWatson
[ Info: Precompiling DrWatson [634d3b9d-ee7a-5ddf-bec9-22491ea816e1]

julia> a = 1; b = 2
2

julia> @strdict a b
Dict{String,Int64} with 2 entries:
  "b" => 2
  "a" => 1

julia> @strdict c=b
Dict{String,Int64} with 1 entry:
  "c = b" => 2

julia> c
2

julia> @strdict a=b
Dict{String,Int64} with 1 entry:
  "a = b" => 2

julia> a
2
```

I think that definitely classifies as a bug.
On the other hand, what do you think, would this be a useful feature?
(e.g. renaming variables)

Datseris:
Hi,

`@strdict c=b`, not sure about that, because it is not the syntax declared by the API of `@dict` (which is what `@savename` does). So I wouldn't even call it a "bug", as, well, what is expected to happen, does happen.

> would this be a useful feature?

I did not understand what is the feature here. That you can re-assign values to variables while calling `@strdict` to them?

JonasIsensee:
Hi, 
I naively hope that 
```
julia> @strdict a=b
Dict{String,Int64} with 1 entry:
  "a" => 2

julia> a
1
```

that the assignment syntax would just use the value of the rhs and use the name of the lhs.

Clearly, this is not implemented but I'm wondering if that would be useful.

Datseris:
Aaaaaaaaaaaaaaaaaaaaaaaaa aaaaaaa

yes, this sounds like an excellent feature!

PR #240, opened 2021-03-25
WIP: allow assignment syntax in @strdict

JonasIsensee:
This PR is a working implementation for #239 

This is still WIP as it needs discussion e.g. what other macros could also benefit from this.


Datseris:
So this allows parsing the following combinations right:
```julia
@dict a b c=some_existing_variable d e=some_other_var
```

sebastianpech:
As this adds macrotools as a dependency, I guess those can be removed:
https://github.com/JuliaDynamics/DrWatson.jl/blob/2b96179159c519f1367433e65dc01900de0a4c31/src/dict_list.jl#L133-L136

sebastianpech:
btw. MacroTools is not >= 1. I don't really care, but I remember a discussion about that ;)

Datseris:
> btw. MacroTools is not >= 1

Nooooooooooooo how can this be!!!!!!!!!!!!!!!!!!!!!!!!!!! There should be a thousand packages that depend on it!!!!!!!!!!!!!!!!!!!

sebastianpech:
> Nooooooooooooo how can this be!!!!!!!!!!!!!!!!!!!!!!!!!!! There should be a thousand packages that depend on it!!!!!!!!!!!!!!!!!!!

Knew it ;) ... Should've let @JonasIsensee sneak the macrotools dependency in

tamasgal:
Oh yeah, the good old not-version-one-yet dilemma and more and more posts on Discord because packages won't upgrade due to pinned versions 🙈

JonasIsensee:
Honestly, semver is a convention. There are never any guarantees.

`MacroTools` is a stable package. It's had a single release in the last 12 months and
from the PR history it does not appear to have had any breaking changes in the last 
2 years.

I vote to just depend on `MacroTools`. If they ever break something, we can still just fix it.

Datseris:
okayyyyyyyyyyyyy okayyyy with a lot of saltiness we can use MacroTools. Jonas, is this correct: https://github.com/JuliaDynamics/DrWatson.jl/pull/240#issuecomment-807382361 ?

JonasIsensee:
> okayyyyyyyyyyyyy okayyyy with a lot of saltiness we can use MacroTools. Jonas, is this correct: [#240 (comment)](https://github.com/JuliaDynamics/DrWatson.jl/pull/240#issuecomment-807382361) ?

Yes, that's correct. :)

src/naming.jl
```diff
@@ -283,10 +284,18 @@ Same as [`@dict`](@ref) but the key type is `String`.
 """
 macro strdict(vars...)
```
Datseris, line 285: So my understanding now is that this works only for `@strdict`? Then it should also be done in `@dict, @ntuple, @savename`.

JonasIsensee: :+1: 
(yes, as I edited above) 
Will do, when I get the chance

Datseris (approved):
cool, also bump minor version in Project.toml please

PR #241, opened 2021-03-26
"dirty"-warning in `gitdescribe` includes `gitpath`

adomasbaliuka:
I use `tag!` to give me information about several repositories that are part of my project. Hence, it's convenient for me to know which one the warning refers to. I think the warning benefits from this additional information. I also think it isn't too long.

This change is very minor and obviously doesn't break anything (`gitpath` is already being interpolated into another warning in the same method). Hence, I do not bother to change other files and adjust version information or the like.


Datseris:
Thanks!

PR #249, opened 2021-05-06
Add Julia project manager link, fix a broken link in docs

thompsonmj:
New to GitHub contributing, new Julia fan, loved the JuliaCon presentation for DrWatson!
Anyway, looking into using DrWatson and noticed a broken link and though the first line in this doc page could use a link out for other noobs like me.


Datseris:
great, thanks!

PR #254, opened 2021-05-22
Add .gitattributes at project initialization

thompsonmj:
Improve cross-platform compatibility and robust handling of binaries, etc with git using .gitattributes.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.0.2"
+version = "2.1.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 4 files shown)


Datseris:
Hi there, I personally do not know what `gitattributes` are and thus I cannot review without you giving a description of what they are and what they do :) Perhaps one of the other devs knows and will review in due time!

thompsonmj:
No problem! The biggest benefit, I think, is `* text=auto` to perform automatic LF normalization for cross-platform projects without assuming contributors have their local git configured to do this automatically. It also automatically checks file content for binaries so git can handle them appropriately. The rest of the entries in the file are taken verbatim from entries [here](https://github.com/alexkaratarakis/gitattributes) that would plausibly show up in a scientific repository. They are explicit rules for how git should handle certain extensions without relying on the detection guessing in the `* text=auto` line, which increases probability of good merges. It would be especially handy if things end up outside their intended directories that are ignored by `.gitignore` to avoid bloating the repository.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.0.3"
+version = "2.1.0"
```
```suggestion
version = "2.0.4"
```

PR #266, opened 2021-07-04
Allow templates in `initialize_project`

### LINKED ISSUE #264: Custom folder structure
I am starting to consider a custom folder structure for a class, and wonder if we can currently customize the structure created by `intiliaze_project`.

I would like to have some sort of DrWatson.jl settings in `.julia/config/drwatson` that would describe the folder structure and the meaning of each folder, e.g. a "src" folder named "code" or a "notebooks" folder named "apps", etc.

I am considering not using DrWatson.jl's initiliaze_project functionality and only use the save/load functionality.

Datseris:
Closes #264

Okay, the idea is pretty simple. A template is just a vector of strings or pair=>vector{string}. Seems pretty clean to me, easy to adopt, easy to change. 

Any ideas/recommendations for a couple more pre-made templates are welcomed.

### DIFF
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -201,7 +201,7 @@ The new project remains activated for you to immidiately add packages.
   is `true` then recursively delete everything in the path and create the project.
 * `git = true` : Make the project a Git repository.
 * `template = DrWatson.DEFAULT_TEMPLATE` : A template containing the folder structure
-  of the project. It should be a vector containing strings (folder) or pairs of `String
+  of the project. It should be a vector containing strings (folders) or pairs of `String
   => Vector{String}`, containg a folder and subfolders (this can be nested further). Example:
   ```
   DEFAULT_TEMPLATE = [
@@ -214,10 +214,11 @@ The new project remains activated for you to immidiately add packages.
     "data" => ["sims", "exp_raw", "exp_pro"],
   ]
   ```
+  Obviously, the default derivative functions of [`projectdir`](@ref), such as `datadir`,
+  have been written with the default template in mind.
 * `placeholder = false` : Add hidden place holder files in each default folder to ensure 
   that project folder structure is maintained when the directory is cloned.
-  Should be used only when `git = true`.
-  Will throw a warning if used with `git = false`
+  Only used when `git = true`.
 """
 function initialize_project(path, name = default_name_from_path(path);
         force = false, readme = true, authors = nothing,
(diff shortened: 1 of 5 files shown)


Datseris:
cc @juliohm

sebastianpech:
@Datseris quite busy at the moment, just skimmed your changes, will have a closer look tomorrow. One thing that came to my mind is, how are we going to treat our directory functions `datadir` etc? Now they aren't useful if one removes the corresponding directories from the template definition.

Datseris:
> Now they aren't useful if one removes the corresponding directories from the template definition.

Yeah, the problem is, we can't, at least not without polluting global scope. It is possible to do some `@eval` hacks, but I don't agree with doing it. The user can use either `projectdir("new",.. )` or create their own versions as we instruct in the docs.

sebastianpech:
I agree with you. I was thinking of creating eg. the `datadir` function only if the directory exists. We could do that during the project activation. Don't know if that's really necessary though.

src/project_setup.jl
```diff
@@ -254,17 +200,31 @@ The new project remains activated for you to immidiately add packages.
...
-    force = false, readme = true, authors = nothing,
-    git = true, placeholder = false)
-
-    placeholder && !git && @warn "The placeholder files are created, but you need to "*
-    "manually commit them to your version control software OR initialize project with git = true"
+        force = false, readme = true, authors = nothing,
+        git = true, placeholder = false, template = DEFAULT_TEMPLATE
+    )
 
+    if git == false; placeholder = false; end
```
sebastianpech, line 227: What's this doing? Did you deliberately remove the warning for placeholder == true, git == false

sebastianpech (approved):
Besides the one comment, the code looks fine.

src/project_setup.jl
Datseris: Yes. Well, as we say in the documentation string, `placeholder` has no reason to be `true` if `git` is `false`. So I just thought of making it `false` always if this is the case.

If this is the wrong way to go let me know and I'll revert.

sebastianpech: I'm good with that. The docstring still contains the info about the warning though.

sebastianpech (approved):

Datseris:
Ok I'll polish, also add some info on the `datadir` business, and merge. If you have a template that you use all the time let me know.

PR #268, opened 2021-07-23
Fix crash of `produce_or_load` when saving file fails.

Klafyvel:
Hi,
I noticed that `produce_or_load` crashes when file saving fails, because of a lonely `*`. This fixes the issue.


Datseris:
hot damn, thanks! Please increment patch version in the Project.toml file, so we can tag a release right after.

Klafyvel:
There you go !

Issue #263, opened 2021-06-28
[FR] function `isdirty(path = projectdir())`.


dingraha:
Often when using DrWatson I'll forget to commit before running something with `produce_or_load`, leading to a dirty git commit/patch. I'd rather have DrWatson refuse to run the function and give me a chance to commit everything and try again. So how about adding a keyword argument like `allow_dirty`, defaulting to `true`, to `produce_or_load` that would throw an error when the repository isn't clean if `allow_dirty==false`?

DrWatson is awesome. Thanks for this package!

Datseris:
Thanks!

I think a better alternative would be to expose a function `isdirty(gitpath)` that returns `true` if it is dirty, and you can decide what you want to do with it. E.g. preface your `produce_or_load` calls with `isdirty(path) && error()`. I don't think that `produce_or_load` should have much to do with `git`, its git-related arguments are just some paths propagated to other functions.

dingraha:
> I think a better alternative would be to expose a function `isdirty(gitpath)` that returns `true` if it is dirty, and you can decide what you want to do with it.

Even better!

dingraha:
Thanks for merging this!

PR #273, opened 2021-07-28
Add `isdirty(gitpath = projectdir())` function

dingraha:
First try at addressing issue #263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2.2.0
+* `isdirty(gitpath = projectdir())` function for checking if `gitpath` points to a dirty Git repository. (#263)
 # 2.1.0
 * `initialize_project` now allows for custom templates.
 # 2.0.4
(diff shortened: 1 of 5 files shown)


Datseris:
Seems pretty good to me! 

* You can use this function inside `tag!` and replace it's raw logic with using the function
* You also need to add it to the docs for users to see it. I'd suggest adding just after `gitpatch` in `docs/src/save.md`

dingraha:
Great, thanks for the feedback.
> * You can use this function inside `tag!` and replace it's raw logic with using the function

Ah, I think you mean that instead of checking if the `patch` returned by `gitpatch` is something other than `""` or `nothing`, `tag!` could use `isdirty`? The worry I have is that `isdirty` as it exists right now can error (if the path passed to it isn't a Git repository, or doesn't exist). That would appear to violate the "never error" rule for DrWatson's saving functions? Would it be better to catch all errors and return `nothing` in `isdirty` (like what's done in `gitdescribe`)?

Datseris:
Yeah I didn't consider the non error part. Then let's leave it as is. Just add the function to the docs and I'll merge.

Datseris:
Thanks @dingraha . I've reduced the amount of new docs a bit by crosslinking things better. Tagging a new release now!

PR #276, opened 2021-08-07
Fix variable name in `findproject`

moble:
If `findproject` doesn't find its requested directory, it tries to issue a warning, but fails because the input variable name was evidently changed from `path` to `dir`.   This fixes it in the warning and the docstring.


Datseris:
Ops sorry about that, my fault! Can you please also increment patch version (third number) in Project.toml so I can immediatelly tag a release?

moble:
Done.

Datseris:
Thanks for the fix!

PR #277, opened 2021-08-11
Fall back on global git config

moble:
Currently, with the default parameters, `initialize_project` makes the project a git repo.  As part of that process, it checks the git config *of the repo* for the options `user.name` and `user.email`.  Unless the repo existed already and had those options set explicitly, these options will not exist, so `initialize_project` just sets them to `"DrWatson"` and `"no@mail"`, respectively.  I'm guessing that the reason for this is to ensure that it can make an initial commit without any issues.

That makes some sense, but I think it makes more sense to check the global config before setting those options.  If they exist in the global config (or already in the local config), they shouldn't be set locally.  The initial commit should still work just fine, using those global settings.


Datseris:
Pretty straightforward, thanks!

PR #281, opened 2021-09-01
Add macro `@produce_or_load` to add source file and line tag

### LINKED ISSUE #136: Macro `@produce_or_load` that also adds the `script` field like `@tagsave`
I've just used `produce_or_load` again in another project, and I was sad to depart with the super-duper-cool functionality of `@tagsave` that adds line number and script file.

But it should be pretty easy to make a `@produce_or_load(args...; kwargs...)` that just takes this information. Given how modular we made `@tagsave` and co., I think this won't be hard.

ulido:
Introduces a macro version of `@produce_or_load` that saves the source
file name and line number of the calling line, similar to `@tagsave`.

Adds tests that check if the appropriate file name and line number was
saved.

This closes #136 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # 2.4.0
 * Add the macro version of `produce_or_load` to enable tagging with the calling source file and line.
-* Refactor `tag!` to avoid code duplication.
 # 2.3.0
 * Enable pass through of kwargs to `wsave` in `produce_or_load`, `tagsave` and `safesave` (to e.g. allow compression in JLD2 files).
 # 2.2.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.4.0
+* Add the macro version of `produce_or_load` to enable tagging with the calling source file and line.
+* Refactor `tag!` to avoid code duplication.
 # 2.3.0
 * Enable pass through of kwargs to `wsave` in `produce_or_load`, `tagsave` and `safesave` (to e.g. allow compression in JLD2 files).
 # 2.2.0
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.3.0"
+version = "2.4.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/docs/src/save.md
+++ b/docs/src/save.md
@@ -56,6 +56,7 @@ In addition, it attempts to minimize computing energy spent on getting a result.
 
 ```@docs
 produce_or_load
+@produce_or_load
 istaggable
 ```
 
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -80,27 +80,6 @@ function produce_or_load(path, c, f::Function;
     end
 end
 
-function scripttag!(d::Dict{K,T}, source::LineNumberNode; gitpath = projectdir()) where {K,T}
-    # This duplicates functionality in `tag!`, for use in `@produce_or_load` when we want to only
-    # include the `script` tag but not enable tagging generally.
-    @assert (Symbol <: K) || (String <: K)
-    if K == Symbol
-        scriptname = :script
-    else
-        scriptname = "script"
-    end
-    if haskey(d, scriptname)
-        @warn "The dictionary already has a key named `script`. We won't "*
-            "overwrite it with the script name."
-    else
-        if !(String <: T)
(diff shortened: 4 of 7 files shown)


ulido:
Please have a good look at my code here, it's my first serious foray into Julia's macros.

Datseris:
Thanks!

1. So if I got this right, at the moment this will add a `:script` field irrespecitively if the keyword `tag` is true or not, right? I'm wondering whether we should check if `tag = true` and only then add a `:script` field. 

2. There is another thing that absolutely must be changed: functions such as `tagsave, safesave, produce_or_load` give the promise that they will never error outside the user's given function. Here however your `@assert` statement will error.

3. at the moment you do not propagate the keyword `gitpath` to `scripttag!`. But of course this needs to be done.

4. I'm wondering... your `scripttag!` is nice. Perhaps we can eliminate code duplication by having all three of `@tag!, @tagsave, @produce_or_load` to use the same function? Would you be willing to do this re-factoring and simplify the code-base?

Datseris:
Oh, also, this should increment minor version, add a changelog entry, and add the macro in the documentation `@docs` code block, just after the normal function `produce_or_load`.

Datseris:
@sebastianpech could you please also review this? You're our macro expert ;)

ulido:
> 1. So if I got this right, at the moment this will add a `:script` field irrespecitively if the keyword `tag` is true or not, right? I'm wondering whether we should check if `tag = true` and only then add a `:script` field.

I'm not so sure about this - the only reason to use `@produce_or_load` over the non-macro version is that it adds the `:script` field. Otherwise they should be functionally identical (unless I've made a mistake).

> 2. There is another thing that absolutely must be changed: functions such as `tagsave, safesave, produce_or_load` give the promise that they will never error outside the user's given function. Here however your `@assert` statement will error.

But `tagsave` calls `tag!` which includes the same `@assert` statement. So if that promise exists, it should be removed there as well.

I think this could also be handled at the method dispatch level, using the function's `where` clause? But this of course would just shift the potential `AssertionError` to a `MethodError`.

> 3. at the moment you do not propagate the keyword `gitpath` to `scripttag!`. But of course this needs to be done.

Thanks for catching that - I'll fix this.

> 4. I'm wondering... your `scripttag!` is nice. Perhaps we can eliminate code duplication by having all three of `@tag!, @tagsave, @produce_or_load` to use the same function? Would you be willing to do this re-factoring and simplify the code-base?

Happy to do this. Do you want this as part of this PR or a subsequent separate refactoring PR?

ulido:
> Oh, also, this should increment minor version, add a changelog entry, and add the macro in the documentation `@docs` code block, just after the normal function `produce_or_load`.

I'll add commits to this end.

ulido:
When refactoring `tag!` I've examined the code. I don't understand this code block here:

https://github.com/JuliaDynamics/DrWatson.jl/blob/22314c4bc15df1bd15585aa06179625b9f2c3eed/src/saving_tools.jl#L211-L222

Essentially, when the `Dict` value type `T` already allows `String`s, it checks `storepatch` and `patch != nothing`. But in the `else` block it only checks `patch != ""`.

Is there a reason for this? This means that `storepatch` is only honored when `String <: T` is `true`.

ulido:
Sorry about the test failures, I'm trying to figure out what went wrong because the tests passed on my machine.

ulido:
Ok, my bad, I was being stupid. Here is what I did:
* I've changed `tag!` now to make it use the new `scripttag!` function.
* I've introduced two new internal helper functions, `checktagtype!` which promotes the value type of `d` to allow Strings if necessary, and `keyname` which change the keys to the key type of `d` (`String` or `Symbol`).
* I've also removed the `@assert` checking the key type and changed it to use method dispatch syntax (i.e. `where {K<:Union{Symbol,String},T}`) - I use the same syntax for `scripttag!` and the other helper functions.
* I've changed the code flow regarding `storepatch` slightly to make more sense - see my post above.

Let me know if you agree with what I did.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.4.0
+* Add the macro version of `produce_or_load` to enable tagging with the calling source file and line.
+* Refactor `tag!` to avoid code duplication.
```
Datseris, line 3: Users don't have to know this :)

Datseris:
great! There is one thing I'm not sure I agree: changing the assertion tag to method dispatch. MethodErrors are not useful for a front end user, because they rarely tell you what's wrong. We could do `@assert L <: Union{String, Symbol} "We only know how to tag dictionaries that have keys that are strings or symbols".` this makes it clear.

CHANGELOG.md
ulido: Now that you say it, that makes sense. :-)

ulido:
Ok, I've changed the method dispatch type check back to the assertion. I've also introduced tests for this.

Datseris:
Thanks @ulido , another quality contribution.

ulido:
Thanks @Datseris !

PR #282, opened 2021-09-02
Automatically convert `String` keys to `Symbol`s in `access`.

### LINKED ISSUE #259: Convert strings to symbols on `accesses` option in savename for composite types
I am trying to use the "accesses" keyword in the function savename to choose only a selection of keys for the file name. While I have no issues with the "ignores" keyword, I can't get it to work with "accesses". This is how I call the function:
`name = savename(Params, "eps"; sigdigits = 3, accesses = ["δ", "phi", "v0"])`
which does give me the following error:
`ERROR: MethodError: no method matching getproperty(::MyParams, ::String)`
(where MyParams is the type of Params, a @with_kw struct).

Whereas 
`name = savename(Params, "eps"; sigdigits = 3, ignores = ["δ", "phi", "v0"])`
does the right thing.

ulido:
To make the `savename` `accesses` keyword syntax easier, allow
automatic conversion of `String` keys to `Symbol`s when working with
containers other than `Dict`s. Explicitly continue to allow `String`
keys for `Dict`s.

Includes tests for `String` key to `Symbol` conversion when handling a
struct, and `String` key pass through when handling a `Dict`.

Closes #259.

### DIFF
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -185,6 +185,10 @@ For example, if `c, c.k1` are `NamedTuple`s then
 access(c, keys...) = access(access(c, keys[1]), Base.tail(keys)...)
 access(c::AbstractDict, key) = getindex(c, key)
 access(c, key) = getproperty(c, key)
+# Automatically convert String keys to Symbols (for structs)...
+access(c, key::AbstractString) = access(c, Symbol(key))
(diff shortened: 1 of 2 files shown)


Datseris:
very nice. clean and straightforward

PR #283, opened 2021-09-04
Allow for OrderedDicts to be passed to dict_list and other similar functions

### LINKED ISSUE #278: Support for OrderedDicts
I have been utilizing OrderedDicts more frequently and have been limited to some of the convenience functions or other things because the functions (e.g. tag! or dict_list) require the input to explicitly be a :Dict{K,T}.  

Since OrderedDicts are the same things as regular Dicts I don't think it should break anything major in the existing code and I just would like to be able to use many of the convenience functions that already exist without having to write a function that basically already exists for a dictionary. 

I think the feature is a simple solution of allowing tag! or similar functions that define the input to be a Dict to also allow OrderedDict. Not sure if that is just an additional method or what.

rs7q5:
Anywhere where functions required the input to be a `::Dict{K,T}` were changed to `::AbstractDict{K,T}` to allow for OrderedDicts.

The only place this was not changed was in the project_setup where it gets the `pkg["name"]` because that returns a Dict.

- Added an optional kwarg `order` argument to `struct2dict` to return an OrderedDict (default is false), where order refers to insertion order. If a set of OrderedDicts is used with `dict_list` an OrderedDict should be returned.
- added a test for using dict_list with an Ordered Dict
- `OrderedDict` from DataStructures is also exported from DrWatson.
- `checktagtype!` will also return a regular Dict.

Fixes #278 at the most basic level.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
 version = "2.4.1"
 
 [deps]
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
--- a/Project.toml
+++ b/Project.toml
@@ -1,9 +1,10 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.4.0"
+version = "2.4.1"
 
 [deps]
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
@@ -23,6 +24,7 @@ julia = "1.0"
 [extras]
 BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
 CSVFiles = "5d742f6a-9f54-50ce-8119-2520741973ca"
+CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
@@ -30,10 +32,6 @@ JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
-CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
 
 [targets]
-test = [
-    "Test", "BSON", "FileIO", "Parameters", "DataFrames",
-    "JLD2", "Statistics", "Dates", "CSVFiles", "CodecZlib"
-]
+test = ["Test", "BSON", "FileIO", "Parameters", "DataFrames", "JLD2", "Statistics", "Dates", "CSVFiles", "CodecZlib"]
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
 version = "2.4.1"
 
 [deps]
-DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
@@ -26,6 +25,7 @@ BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
 CSVFiles = "5d742f6a-9f54-50ce-8119-2520741973ca"
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
@@ -34,4 +34,4 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "BSON", "FileIO", "Parameters", "DataFrames", "JLD2", "Statistics", "Dates", "CSVFiles", "CodecZlib"]
+test = ["Test", "BSON", "FileIO", "Parameters", "DataFrames", "JLD2", "Statistics", "Dates", "CSVFiles", "CodecZlib", "DataStructures"]
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
 version = "2.4.1"
 
 [deps]
-DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -60,7 +60,7 @@ See also [`parse_savename`](@ref) and [`@savename`](@ref).
   called with its default arguments (so customization here is possible only
   by rolling your own container type). Containers leading to empty `savename`
   are skipped.
-* `equals = "="` : Connector between name and value. Can be useful to modify for 
+* `equals = "="` : Connector between name and value. Can be useful to modify for
   adding space `" = "`.
 
 ## Examples
@@ -271,7 +271,7 @@ end
 
 """
     esc_dict_expr_from_vars(vars)
-Transform a `Tuple` of `Symbol` and assignments (`a=b`) 
+Transform a `Tuple` of `Symbol` and assignments (`a=b`)
 into a dictionary where each `Symbol` in `vars`
 defines a key-value pair. The value is obtained by evaluating the `Symbol` in
 the macro calling environment.
@@ -356,10 +356,10 @@ ntuple2dict(nt::NamedTuple) = Dict(k => nt[k] for k in keys(nt))
 Convert a dictionary (with `Symbol` or `String` as key type) to
 a `NamedTuple`.
 """
-function dict2ntuple(dict::Dict{String, T}) where T
+function dict2ntuple(dict::AbstractDict{String, T}) where T
     NamedTuple{Tuple(Symbol.(keys(dict)))}(values(dict))
 end
-function dict2ntuple(dict::Dict{Symbol, T}) where T
+function dict2ntuple(dict::AbstractDict{Symbol, T}) where T
     NamedTuple{Tuple(keys(dict))}(values(dict))
 end
 
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -369,14 +369,13 @@ Change a dictionary with key type `Symbol` to have key type `String`.
 """
 tostringdict(::Type{DT},d) where {DT<:AbstractDict} = DT(zip(String.(keys(d)), values(d)))
 tostringdict(d) = tostringdict(Dict,d)
-#tostringdict(d) = Dict(zip(String.(keys(d)), values(d)))
 
 """
     tosymboldict(d)
 Change a dictionary with key type `String` to have key type `Symbol`.
 """
 tosymboldict(::Type{DT},d) where {DT<:AbstractDict} = DT(zip(Symbol.(keys(d)), values(d)))
-tosymboldict(d) = tosymboldict(Dict,d)#Dict(zip(Symbol.(keys(d)), values(d)))
+tosymboldict(d) = tosymboldict(Dict,d)
 
 """
     parse_savename(filename::AbstractString; kwargs...)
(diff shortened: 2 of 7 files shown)


Datseris (commented):
Thanks for the pull request. in DrWatson we try to keep the amount of dependencies as minimal as possible, but unfortunately I see that this adds a dependency on `DataStructures`. I am not sure why this is the case. Why do we need it? If you just replace all `Dict` with `AbstractDict`, shouldn't it work without the need for `DataStructures`?

Datseris:
Ah now I see why the `DataStructures`, so that struct2dict can return an OrderedDict . Sorry, but I don't think this justifies the dependency :( a user can just do `OrderedDict(structu2dict(...))`

rs7q5:
I don't know if stuct2dict and then wrapping in ordered dict would maintain insertion order. I could be totally wrong there. In theory I think you could just import only OrderedDict. But I see your point. The only potential issue one would have would be when anything is returned, it would not return in the same way (i.e. dict_list). I also wasn't sure if I still needed it to run the testset of an ordered dict, which may of been why I also included it.

Datseris:
But if you care about ordering, why do you need to convert a struct to a dict? A struct also has ordering. In general, can you justify what's the benefit of the ordering?

rs7q5:
Besides deterministically being able to iterate over the dictionary? Also, insertion order and sorting by key are two different things. I don't think Dict takes into account any of that consistently. I have not tested or looked into it though. I did test your example and it does not produce the same results, which after looking at it makes sense because Order is insertion order not key order.  I don't know if you would ever be able to recover insertion order if you were to save and load in another file (i.e. to run something else or analyze results, etc.).  MWE of this at the end though.

Personally, I have found insertion order important, particularly when analyzing experimental results (I convert experimental data from LabVIEW using [TDSMReader.jl](https://github.com/DanBerge/TDMSReader.jl)), which uses OrderedDicts. I understand this is a specific use, but not having to think about if something is in a large structure of variables and knowing it is roughly in a particular location and should be consistently is one of those things that I find spending less time needing to check when coding. 

Thinking about it though, I think having the extra features like :gitcommit, etc. always at the end of the file (or the option to do so would be convenient to know if they are there, that is where they are located. (I do understand that is a separate issue though).  

Also, if it helps it looks like there is the package [OrderedCollections.jl](https://github.com/JuliaCollections/OrderedCollections.jl), which includes `OrderedDicts` and `LittleDicts`. This is split from `DataStructures` (which also includes `defaultDicts` and `defaultOrderedDicts`), I don't know if that impacts the dependency issue since it's still dependent. Could maybe reduce compile time if that is also an issue. One could also just import OrderedDicts from either of those. 

**MWE of struct2dict with ordering**
```
using DataStructures
function struct2dict_new(s;order=false)
    if order
        OrderedDict(x => getfield(s, x) for x in fieldnames(typeof(s)))
    else
        Dict(x => getfield(s, x) for x in fieldnames(typeof(s)))
    end
end

struct StructA
    c::Float64
    a::Float64
    d::Float64
end

    d = StructA(1,2,4)
    var1 = struct2dict_new(d)
    var2 = struct2dict_new(d;order=true)
    var3 = OrderedDict(struct2dict_new(d))
```

JonasIsensee:
Hi!Thanks for your contribution. I haven't had the time to look at your changes in detail but I just wanted to weigh in on the DataStructures discussion. 1) DataStructures is a very common and very stable dependency. It is already a dependency implicitly through JLD2.2) if it is truly just for a single function, then you could consider changing the signatures :```struct2dict(DT::Type{<:AbstactDict}, args...) =... struct2dict(args...) = struct2dict(Dict, args...) ```That wouldn't do anything about the defaults but it would allow you to use OrderedDicts without a dependency.

rs7q5:
I think the defaults would only apply to a specific project, not necessarily generally. It may be beneficial in the future, but I think the use or optional use of DrWatson would be very convenient, at least for my use. I was just mentioning other potentially helpful functions.

I only put OrderedDicts in a single function or two (I used in dict_list too if an OrderedDict was provided) because I don't currently have the time to play test all of it.

Also I appreciate your input on this matter, I don't know much about the intricacies of a lot of things at this level of coding as I'm not a programmer by trade.

Datseris:
Thanks for weighing in!

>  It is already a dependency implicitly through JLD2.

JLD2 is _not_ a direct dependency of DrWatson. Only FileIO is. 

> That wouldn't do anything about the defaults but it would allow you to use OrderedDicts without a dependency.

How would that work? The user would still need to write a method for `struct2dict(WhateverDict, args...)`. This is the same as writing their own function anyways, and doesn't really justify putting this functionality here.

My main concerns are these:
1. Originally these conversion functions were just convenience so that you can pass something to `savename`.
2. The added function with `OrderedDict` is a bit odd. It is a conversion to a thing that doesn't come from DrWatson, but from another package. It is weird to me that DrWatson should offer this functionality, instead of the package that actually defines `OrderedDict`.
3. It adds a dependency.

@rs7q5 can you please do me a favor and change this PR to _just_ doing the `Dict -> AbstractDict` change? This solves the original issue cited. Whether DrWatson should get a dependency on `DataStructures` and do conversions for things it doesn't really know is a different story that needs to be discussed in a different PR. Personally I'm not convinced the dependency is worth it, but perhaps Jonas is. The question @JonasIsensee is not whether the dependency is stable or not, it is whether it is worth it for this simple function.

JonasIsensee:
> JLD2 is not a direct dependency of DrWatson. Only FileIO is.

ok, nevermind then.

> How would that work? The user would still need to write a method for struct2dict(WhateverDict, args...). This is the same as writing their own function anyways, and doesn't really justify putting this functionality here.

Since `struct2dict` is so extremely simple, here's a full example:
The function body of `struct2dict` inserts elements into the dictionary 
in the correct order. Thus, it is not necessary to implement your own method.
All you need to do, is pass a type `<: AbstractDict` that implements the
regular dictionary constructor. 
```
julia> function struct2dict(::Type{DT}, s) where {DT <: AbstractDict}
           DT(x => getfield(s, x) for x in fieldnames(typeof(s)))
       end
struct2dict (generic function with 1 method)

julia> struct2dict(s) = struct2dict(Dict, s)
struct2dict (generic function with 2 methods)

julia> struct TestStruct; y::Int; z::Float64; x::String; end

julia> t = TestStruct(1, 2.0, "3")
TestStruct(1, 2.0, "3")

julia> struct2dict(t) # here the ordering also happens to be correct
Dict{Symbol, Any} with 3 entries:
  :y => 1
  :z => 2.0
  :x => "3"

julia> using DataStructures

julia> struct2dict(OrderedDict, t) # here the ordering is guaranteed to be the same as in `fieldnames(typeof(t))`
OrderedDict{Symbol, Any} with 3 entries:
  :y => 1
  :z => 2.0
  :x => "3"
  ```

So, the point I'm trying to make:
In some cases, it would be super easy to open up the API in a way to allow julian duck-typing. (As you can see from my example, this does not require additional dependencies)

**Comment by Datseris** [comment]
Well, that's lovely then, because it seems to solve both the dependency problem and actually allowing our nice user to be satisfied and have a conversion to OrderedDict.

**Comment by rs7q5** [comment]
The latest commit should have removed the DataStructures dependency. I also went ahead and made the suggestion by @JonasIsensee  on the struct2dict thing.

File: src/saving_tools.jl [hunk 1]
```diff
@@ -225,37 +225,38 @@ function tag!(d::Dict{K,T}; gitpath = projectdir(), storepatch = true, force = f
...
-    checktagtype!(d::Dict{K,T}) where {K<:Union{Symbol,String},T}
+    checktagtype!(d::AbstractDict{K,T}) where {K<:Union{Symbol,String},T}
 
 Check if the value type of `d` allows `String` and promote it to do so if not.
+Currently will accept an abstract dict, but will return a dict.
 """
-function checktagtype!(d::Dict{K,T}) where {K<:Union{Symbol,String},T}
+function checktagtype!(d::AbstractDict{K,T}) where {K<:Union{Symbol,String},T}
     if !(String <: T)
         d = Dict{K, promote_type(T, String)}(d)
```
Datseris, line 247: Shouldn't this line also be changed? Instead of `Dict` to use `typeof(d)` ?

Datseris: I mean if the value types already accept `String` it won't change to Dict, but otherwise it will? Sounds weird.

Datseris: It should either always change to Dict or never, for stability reasons.

Datseris:
@rs7q5 

1. Why is CodeZLib in the test suite now?
2. Please re-add the tests that use `OrderedDict` to make sure that the functionality actually works. Having a dependency in the test suite is no problem :)

src/saving_tools.jl
rs7q5: It should be. I will change that.

rs7q5:
I'm not sure how CodeZLib is in the test suite. Where is it? I didn't add it. 

Do we just need tests for struct2dict and checktagtype or would it be run through istaggable? 

dict_list should now accept any dicts but return a regular dict. Unless you think it should be changed similarly as well. I wasn't sure what else would break, so I left it alone.

Project.toml
```diff
@@ -23,17 +23,14 @@ julia = "1.0"
 [extras]
 BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
 CSVFiles = "5d742f6a-9f54-50ce-8119-2520741973ca"
+CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
```
Datseris, line 26: Here it says that CodecZlib was added in this PR.

Datseris:
I think adding the following tests would be enough:

* `tagsave` works with OrderedDict (this also chacks `istaggable`)
* struct2dict with ORderedDict returns an ordered dict

It is fine that `dict_list` returns standard `Dict`. It is not meaningful to have ordered keys for `dict_list` given its target goal.

Project.toml
rs7q5: CodecZlib is for TranscodingStreaming or something, no idea why it is there, a quick look at the documentation shows it may be used for compression, so this would not be part of my PR. It looks like compared to my Project.toml file, the package just got moved earlier in the [extras] list compared to what DrWatson 2.4.0 has.

Datseris: Ooooh sorry about that my fault. Yeap it is added in another PR, I forgot... apologies.

src/saving_tools.jl
rs7q5: Is there a current test case where this if statement does get triggered? I haven't been able to find one to test this correctly.

rs7q5: Okay so I think this is now fixed. using only `typeof` does not work because it would still return the `{K,V}`  is still returned as well. Through searching I did find a function that does return the concrete type (e.g. Dict or OrderedDict, etc.  

The function is [`get_rawtype`](https://discourse.julialang.org/t/retrieve-the-type-of-abstractdict-without-parameters-from-a-concrete-dictionary-type/67567/4) from a discussion thread. The only thing is if `parentmodule` will still work even though `DataStructures` is not a dependency. My guess though is if you have to load `DataStructures` in yourself Julia should be aware of that module, since you can't use other DictTypes unless you imported it's parent module. 

I could not figure out a better way to do this. It passes the tests of sending in an OrderedDict to tagsave, but it does return a regular dict. Now, I'm not sure why because I can't find a place where it is explicitly converting it to a regular dict, but I do have a test in there.  Honestly I am not overly concerned with that behavior at the moment because if one really wanted to add tag save, etc and maintain an order they would probably have to overload `wsave` anyway additionally, `collect_results` does not return them in a set order anyway (similar to your thoughts on `dict_list`). 

As a sanity check though I have included a test to `checktagtype!` explicitly. All tests are currently in `stools_tests.jl` because Most of the usage of `OrderedDicts` is a specific tool and could technically be branched off into a separate test file if you think that would add clarity. 

Now the only thing remaining is the dependency of `DataStructures` is still in my [`deps`] section of the Project.toml. I could not run the test cases without it. Now I am on Julia 1.5.2 and there does seem to be a change to the way test-specific dependencies are added, but I don't think that is the issue here. I've tried just about everything. 

I have submitted a new pull-request though for review, but I think just about everything should be there beyond the `DataStructures` issue. I'd be happy to fix it if you have any ideas on how to get it to do so.

rs7q5:
See above for update on this commit https://github.com/JuliaDynamics/DrWatson.jl/pull/283#discussion_r702500810.

Project.toml
```diff
@@ -1,9 +1,10 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.4.0"
+version = "2.4.1"
 
 [deps]
+DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
```
Datseris, line 7: This must be in `[extras]`, not in `[deps]`

rs7q5: Correct, I was having issues getting it there, but apparently I just needed to add it to the target list. This should all actually be fixed now

src/naming.jl
```diff
@@ -356,24 +356,27 @@ ntuple2dict(nt::NamedTuple) = Dict(k => nt[k] for k in keys(nt))
...
 end
 
 """
     tostringdict(d)
 Change a dictionary with key type `Symbol` to have key type `String`.
 """
-tostringdict(d) = Dict(zip(String.(keys(d)), values(d)))
+tostringdict(::Type{DT},d) where {DT<:AbstractDict} = DT(zip(String.(keys(d)), values(d)))
+tostringdict(d) = tostringdict(Dict,d)
+#tostringdict(d) = Dict(zip(String.(keys(d)), values(d)))
```
Datseris, line 372: Please remove previous code. If you've changed the code and the new code works and is better, there is no reason to keep code commented. Commented code is confusing.

src/naming.jl
```diff
@@ -356,24 +356,27 @@ ntuple2dict(nt::NamedTuple) = Dict(k => nt[k] for k in keys(nt))
...
 
 """
     tosymboldict(d)
 Change a dictionary with key type `String` to have key type `Symbol`.
 """
-tosymboldict(d) = Dict(zip(Symbol.(keys(d)), values(d)))
+tosymboldict(::Type{DT},d) where {DT<:AbstractDict} = DT(zip(Symbol.(keys(d)), values(d)))
+tosymboldict(d) = tosymboldict(Dict,d)#Dict(zip(Symbol.(keys(d)), values(d)))
```
Datseris, line 379: Same. Here remove commented code.

src/naming.jl
```diff
@@ -407,7 +410,7 @@ function parse_savename(filename::AbstractString;
     last_dot = findlast(".",savename_part)
     if last_dot == nothing || last_eq > last_dot
         # if no dot is after the last "="
-        # there is no suffix
+        # there is no suffixf
```
Datseris, line 413: Where did this change come from? suffix -> suffixf? I guess it's a typo that should be reversed?

src/saving_tools.jl
```diff
@@ -329,16 +338,18 @@ istaggable(x) = x isa AbstractDict
 
 
 """
-    struct2dict(s) -> d
+    struct2dict(::Type{DT},s) -> d where {DT<: AbstractDict}
```
Datseris, line 341:
```suggestion
    struct2dict([type = Dict, ] s) -> d
```

Datseris (commented):
Thanks, this is almost there. Only some minor comments remaining and then I can merge away :)

src/naming.jl
rs7q5: I knew I was forgetting something. thanks for catching that. And I agree, commented code is confusing.

src/saving_tools.jl
rs7q5: I went ahead and made this change, but just out of curiosity, does the notation with the brackets imply just another method(as opposed to obviously kwargs options at the end of a function). I understood what it does, but wasn't sure when to use that type of notation.

Datseris: the bracket notation inside documentation strings means optional _positional_ arguments. Of course for Julia this also coincides with different methods.

PR #285, opened 2021-09-06
update defaults for keyword arguments of `savename`

### LINKED ISSUE #284: `savename` documentation
Hi,
I just noticed that the default values for the `savename` keywords that are given in the docs:
https://github.com/JuliaDynamics/DrWatson.jl/blob/953888eb7c8782ae1278601a0fc9d563ed7803c7/src/naming.jl#L37
do not match the real default values:
https://github.com/JuliaDynamics/DrWatson.jl/blob/953888eb7c8782ae1278601a0fc9d563ed7803c7/src/naming.jl#L87-L89

I would like to argue that the documented default values are the more sensible choice, because I just found out, that for `h = [1.0, 0.1, 0.01, 0.001, 0.0001, 1.0e-5, 1.0e-6]` we get 
```julia
julia> savename.(Dict.(:h.=>hs))
7-element Vector{String}:
 "h=1.0"
 "h=0.1"
 "h=0.01"
 "h=0.001"
 "h=0.0"
 "h=0.0"
 "h=0.0"
```
whereas with `sigdigits=3` we get
```julia
julia> savename.(Dict.(:h.=>hs), sigdigits=3)
7-element Vector{String}:
 "h=1.0"
 "h=0.1"
 "h=0.01"
 "h=0.001"
 "h=0.0001"
 "h=1e-5"
 "h=1e-6"
```
but I am aware that similar examples can be found for every possible default values for `digits`/`sigdigits`.
Should I open a PR to change the docstring or the default values?

fkastner:
This closes #284.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 2.4.1
-* `savename`'s default options now have `sigdigits = 3` instead of `digits = 3`. (#284)
+* `savename`'s default options now have `sigdigits = 3` instead of `digits = 3` as stated in the documentation string. This was supposed to happen already since 2.0 but did not because of a bug. (#284)
 # 2.4.0
 * Add the macro version of `produce_or_load` to enable tagging with the calling source file and line.
 # 2.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# 2.4.1
+* `savename`'s default options now have `sigdigits = 3` instead of `digits = 3`. (#284)
 # 2.4.0
 * Add the macro version of `produce_or_load` to enable tagging with the calling source file and line.
 # 2.3.0
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.4.0"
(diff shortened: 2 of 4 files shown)


Datseris:
@fkastner please do add an entry in the CHANGELOG.md about this. Technically it is a breaking change but since it is a bugfix it doesn't count as breaking. We should still let users know somewhere.

fkastner:
It is actually already documented as such for the 2.0 release in the changelog. :)
But I will add it again. Should I also bump the version number to 2.4.1?

Datseris:
Yes please

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# 2.4.1
+* `savename`'s default options now have `sigdigits = 3` instead of `digits = 3`. (#284)
```
Datseris, line 2:
```suggestion
* `savename`'s default options now have `sigdigits = 3` instead of `digits = 3` as stated in the documentation string. This was supposed to happen already since 2.0 but did not because of a bug. (#284)
```

fkastner:
I will look at the test failures now. It seems like `sigdigits` currently always overrides `digits`.

fkastner:
Locally all tests pass now.
Some remarks:
1. I had to modify one test to the new default convention.
2. With this fix the behaviour when specifying *both* `digits` *and* `sigdigits` changed. Where before `sigdigits` would take precedence, now `digits` 'wins'. Since I hope nobody relies on the current behaviour ([relevant xkcd](https://xkcd.com/1172/)) this should be fine.

Issue #150, opened 2020-05-12
Interacting with data


JonasIsensee:
I've been working a lot with `produce_or_load` recently.
I think it's a great tool, especially when things are still under development and it saves a lot of time when having to make tiny adjustments to plots or other places.

But of course all data becomes obsolete once the simulation code is updated.
In fact, in that case I tend to trick myself by forgetting that only the results are loaded and wonder why my results are not changing.
Every time that happens I have to navigate to my data folder and select the old files and delete them by hand.

To be honest I don't have a thought out solution to this.
Different ideas floating around my head are:
- Including some kind of hash of the code / callstack in the filenames?
- Provide an interface similar to `collect_results` to view the existing files and manipulate/delete them
- Provide a `reset` function that deletes / archives all files matching the naming pattern of the corresponding `produce_or_load` function

Datseris:
This was what motivated me to add the `force` keyword to `produce_or_load`. This of course doesn't solve the entire problem: in my use case, the total combination of parameters and data were small, and so was the codebase, and thus I could actually remember when I had to force something to be re-evaluated...

What I wonder though, is can this problem even be solved?

What you want here is to re-evaluate the computations that produce the data. **But**, you don't want this to happen based on simply whether the file exists or not; this is what `produce_or_load` already does. You want the re-evaluation to happen once your code base has "progressed enough". I guess the hard part is quantifying deterministically when this happens.

Checking the `git` commit is not a good alternative, because it doesn't deterministically guarantee that the function used in `produce_or_load` has changed in the mean time. You may just added more totally irrelevant new scripts and committed those.

Datseris:
The `reset` idea sounds nice, and it could in fact be an additional dispatch type of `force`. But can you explain how you would do it? I haven't fully understood how this would work. Can you provide a toy example?

JonasIsensee:
```
function expensive_computation(p)
    sleep(1)
    return Dict(:result => p.a * rand(10))
end
map(1:10) do a
    produce_or_load((a=a,), expensive_computation; kwargs...)
end
```

Ok suppose we have the above use-case and I am to change something about the definition of `expensive_computation` that makes all old results useless.
(Or maybe just some??)

Then we could implement a function `reset_data` or so, that takes
`typeof((a=a,))` and `kwargs` to compute the path of all data files as well as a regexp to match filenames using the parameterset.
In this case this would match (and delete) all filename of the form `a=XXX.bson` where `XXX` is some Integer.

[...]

PR #286, opened 2021-09-07
Introduce `update` argument to `collect_results!` to sync results files.

ulido:
I often re-run a computation or data analysis that results in some updated data files. But this isn't propagated through to the results database by `collect_results!` because that only looks for new files but leaves entries for files that have already existed but are now changed alone. The only way to trigger an update is to delete the results database. However, I often run some moderately lengthy computation in `collect_results!` using the `special_list` argument and want to avoid doing that for the whole collection of data files.

This PR adds an `update` keyword argument to `collect_results!` (defaulting to `false`). If `true` it actively syncs an existing results database with the raw data file collection. In particular, when it encounters files whose `mtime` is younger than the result database's `mtime`, it removes the corresponding entry from the database and recreates it using the newer file's data. When it encounters an entry for a file that does not exist anymore it removes that entry.

A few points:
* This is loosely connected with the discussion in #150, about having a way for `produce_or_load` to check if a file needs to be updated because the code changed.
* This might stray into "give the user enough rope to hang themselves" territory. I'm curious what you think about it, but it is definitely useful for me. But I do understand if you don't want to merge this because of data safety concerns.
* I haven't yet added a dire warning to the documentation of the function that "this can mess with your results databases, so do not activate if you do not want that!", but something like this should probably go in.
* I only tested this on macos, but it should work the same on Linux. I don't have a way to test it on Windows, where `mtime` might not work as expected. I wrote tests for it, so it should be revealed during the CI run. 
* Alternatively, the update check could be done by using file hashes instead of `mtime`, but this would be slightly more costly, albeit more portable across platforms.

### DIFF
--- a/src/result_collection.jl
+++ b/src/result_collection.jl
@@ -39,6 +39,8 @@ See also [`collect_results`](@ref).
 * `rpath = nothing` : If not `nothing` stores `relpath(file,rpath)` of result-files
   in `df`. By default the absolute path is used.
 * `verbose = true` : Print (using `@info`) information about the process.
+* `update = false` : Update data from modified files and remove entries for deleted
+  files.
 * `white_list` : List of keys to use from result file. By default
   uses all keys from all loaded result-files.
 * `black_list = [:gitcommit, :gitpatch, :script]`: List of keys not to include from result-file.
@@ -75,15 +77,18 @@ function collect_results!(filename, folder;
     subfolders = false,
     rpath = nothing,
     verbose = true,
+    update = false,
     newfile = false, # keyword only for defining collect_results without !
     kwargs...)
 
     if newfile || !isfile(filename)
         !newfile && verbose && @info "Starting a new result collection..."
         df = DataFrames.DataFrame()
+        mtime_df = 0.0
     else
         verbose && @info "Loading existing result collection..."
         df = wload(filename)["df"]
+        mtime_df = mtime(filename)
     end
     @info "Scanning folder $folder for result files."
 
@@ -99,23 +104,41 @@ function collect_results!(filename, folder;
     end
 
     n = 0 # new entries added
+    u = 0 # entries updated
     existing_files = "path" in string.(names(df)) ? df[:,:path] : ()
     for file ∈ allfiles
         is_valid_file(file, valid_filetypes) || continue
         # maybe use relative path
         file = rpath === nothing ? file : relpath(file, rpath)
+        replace_entry = false
         #already added?
-        file ∈ existing_files && continue
-
+        if file ∈ existing_files
+            # older than df?
+            (!update || (mtime_df > mtime(file))) && continue
+            replace_entry = true
+        end
+            
         data = rpath === nothing ? wload(file) : wload(joinpath(rpath, file))
         df_new = to_data_row(data, file; kwargs...)
         #add filename
         df_new[!, :path] .= file
-
+        if replace_entry
+            # Delete the row with the old data
+            delete!(df, findfirst((x)->(x.path == file), eachrow(df)))
+            u += 1
+        else
+            n += 1
+        end
         df = merge_dataframes!(df, df_new)
-        n += 1
     end
-    verbose && @info "Added $n entries."
+    if update
+        # Delete entries with nonexisting files.
+        idx = findall((x)->(!isfile(x.path)), eachrow(df))
+        delete!(df, idx)
+        verbose && @info "Added $n entries. Updated $u entries. Deleted $(length(idx)) entries."
+    else
+        verbose && @info "Added $n entries."
+    end
     !newfile && wsave(filename, Dict("df" => df))
     return df
 end
--- a/src/result_collection.jl
+++ b/src/result_collection.jl
@@ -84,11 +84,19 @@ function collect_results!(filename, folder;
     if newfile || !isfile(filename)
         !newfile && verbose && @info "Starting a new result collection..."
         df = DataFrames.DataFrame()
-        mtime_df = 0.0
+        mtime_df = nothing
     else
         verbose && @info "Loading existing result collection..."
         df = wload(filename)["df"]
-        mtime_df = mtime(filename)
+        # Check if we have pre-recorded mtimes (if not this could be because of an old results database).
+        if "_mtime" ∈ names(df)
+            mtime_df = nothing
+        else
+            @warn "Update of existing results collection requested, but no previously recorded modification time found. Will proceed by comparing with the database modification time, but this may miss modifications if the database is newer than the modified files."
+            mtime_df = mtime(filename)
+            # Insert a new column to hold mtime information
+            df[!,:_mtime] .= NaN
+        end
     end
     @info "Scanning folder $folder for result files."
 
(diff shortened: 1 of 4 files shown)


test/update_results_tests.jl
```diff
@@ -132,6 +132,39 @@ subfolders = true, special_list=special_list, black_list = black_list)
 @test sort(names(cres3)) == sort(names(cres2))
 @test size(cres3) == size(cres2)
 
+###############################################################################
+#                           test updating feature                             #
+###############################################################################
+
+folder = datadir("update_test")
+mkdir(folder)
```
JonasIsensee, line 140: I'm not a big fan of these persistent files / folders.

This works just fine in CI (because tests only run once) but when testing locally, everything breaks once you rerun individual tests.

@Datseris What is your opinion on this?

src/result_collection.jl
```diff
@@ -75,15 +77,18 @@ function collect_results!(filename, folder;
...
     kwargs...)
 
     if newfile || !isfile(filename)
         !newfile && verbose && @info "Starting a new result collection..."
         df = DataFrames.DataFrame()
+        mtime_df = 0.0
     else
         verbose && @info "Loading existing result collection..."
         df = wload(filename)["df"]
+        mtime_df = mtime(filename)
```
JonasIsensee, line 91: Hm, this appears to be rather brittle to me.

Suppose you 
- add and modify files
- then call `collect_results` without the `update` kw
- then remember and call it again with `update`

In this scenario, your updated files never make it into the `DataFrame`.
It may be more robust to attach `mtime` attributes for all data files individually.

JonasIsensee (commented):
Hi @ulido,

thank you for this. I've been wanting a feature like this for a long time but never made the effort of implementing it myself.

test/update_results_tests.jl
ulido: I agree. I went with what was done in other tests in the same file, but it probably would be better to use temporary directories with some cleanup code that is run regardless if tests fail. I can look into it.

src/result_collection.jl
ulido: This is a good point. I considered adding an `mtime` field (or even better a hash of the file, that'd be even more reliable). But this would not work for files created by `collect_results!` before this was implemented. And I wanted the current behaviour to be preserved for `update=false`.

I guess we could check against the database's `mtime` if no `mtime` or `hash` field is there, issuing a warning and then creating the field?

Datseris:
(Jonas, I'm really busy with submitting a proposal. Thanks for doing a review. I'll leave the merging and tagging a new version here for you. thanks Ulido. Please don't forget the changelog stuff)

test/update_results_tests.jl
ulido: Ok, I've pushed a commit that changes the tests to run in a temp directory.

ulido:
In the latest commit I've implemented checking against individual file `mtime`s. I needed to update the logic in `collect_results!` to accommodate old results files without the `_mtime` field (issuing a warning that we are now checking against the `mtime` of the database).

This also required some updating of tests that assume that a results collection has a certain number of columns - I've incremented that by one to account for the new `_mtime` column.

src/result_collection.jl
JonasIsensee: I'm wondering, whether it makes sense to include the `mtime` or `hash` or whatever in the *user-facing* `DataFrame`.
One could alternatively carry it around separately. That would keep the field more of an implementation detail .I'm not sure that's useful though.

ulido: I've now put the mtime it in the user-facing `DataFrame`. But we could put it as a `Dict` with `path` => `mtime` under an additional `mtime` key in the saved "results_*.jld2" file instead of the `DataFrame`. That might even simplify the logic.

JonasIsensee: I think I'd prefer that

ulido:
> I think I'd prefer that
>
> To elaborate:
> This mtime thing is mostly an implementation detail to reduce computation cost.
> It shouldn't convolute the users aggegrated data.

Agree! I've now changed it to a `Dict` under the `mtime` key in the "results_*.jld2" file.

src/result_collection.jl
```diff
@@ -75,15 +77,29 @@ function collect_results!(filename, folder;
...
         verbose && @info "Loading existing result collection..."
-        df = wload(filename)["df"]
+        data = wload(filename)
+        df = data["df"]
+        # Check if we have pre-recorded mtimes (if not this could be because of an old results database).
+        if "mtime" ∈ keys(data)
+            mtimes = data["mtime"]
+            mtime_df = nothing
+        else
+            @warn "Update of existing results collection requested, but no previously recorded modification time found. Will proceed by comparing with the database modification time, but this may miss modifications if the database is newer than the modified files."
```
JonasIsensee, line 98: This update warning appears to be thrown even if `update = false` which appears to contradict the statement in the warning.

ulido:
If you're happy with this, I'll now add a changelog entry and increase the minor version?

JonasIsensee:
Hi @ulido,
I am happy with most of the changes.
However, I think I may have found a potential bug that I am testing just now.

JonasIsensee (commented):
Ok, so one last thing.

Given an existing (old) results file, calling
`collect_results(folder; update = false)`
yields a corrupted `mtimes` dictionary that is missing entries of non-new files.
This errors on the next attempt to update.

In terms of your tests:
```
cres_before = collect_results!(folder; update = true)
wsave(joinpath(dirname(folder), "results_$(basename(folder)).jld2"), Dict("df" => cres_before))

d = Dict("idx" => :to_modify, "b" => "modified_value")
DrWatson.wsave(fname_modify, d)
rm(fname_delete)

cres_after = collect_results!(folder; update = false)
cres_after = collect_results!(folder; update = true)	
```

ulido:
> Ok, so one last thing.
> 
> Given an existing (old) results file, calling
> `collect_results(folder; update = false)`
> yields a corrupted `mtimes` dictionary that is missing entries of non-new files.
> This errors on the next attempt to update.
> 
> In terms of your tests:
> 
> ```
> cres_before = collect_results!(folder; update = true)
> wsave(joinpath(dirname(folder), "results_$(basename(folder)).jld2"), Dict("df" => cres_before))
> 
> d = Dict("idx" => :to_modify, "b" => "modified_value")
> DrWatson.wsave(fname_modify, d)
> rm(fname_delete)
> 
> cres_after = collect_results!(folder; update = false)
> cres_after = collect_results!(folder; update = true)	# this errors
> ```

Thanks for finding this! There is no codepath to update mtimes of files that don't need updating but it would an easy fix to add this.

> I think, the most consistent workaround would be to not upgrade old files.

You mean, to refuse to update result collection files that do not have a `mtime` entry in the dictionary. I agree, that's probably most consistent. I can change the code to do this. Do you think It'd be best to `@warn` about this and then fall back to the behavior of `upgrade = false` or throw an error?

src/result_collection.jl
ulido: Oops, that just needs a test for `update`.

JonasIsensee:
> Thanks for finding this! There is no codepath to update mtimes of files that don't need updating but it would an easy fix to add this.

The point is, that then the warning would not be thrown again and a (possibly) wrong mtime would be added to the file.

> > I think, the most consistent workaround would be to not upgrade old files.
> 
> You mean, to refuse to update result collection files that do not have a `mtime` entry in the dictionary. I agree, that's probably most consistent. I can change the code to do this. Do you think It'd be best to `@warn` about this and then fall back to the behavior of `upgrade = false` or throw an error?

So the cases I imagine:

1) no `mtime` entry + `update=false` → current/old behaviour (maybe a warning) DO NOT add an `mtime` entry
2) no `mtime` entry + `update=true` → error, tell to recompute ( this is safer, so users don't actually create wrong time stamps and continue to work with old data)
3) with `mtime` entry → new behaviour
4) newfiles always receive an `mtime` entry

To me, this seems like the most consistent / least dangerous option.

ulido:
That makes sense, and I've implemented this in the latest commit.

JonasIsensee (approved):

JonasIsensee:
Hi @ulido,
thank you so much for your patience.
I believe, we've arrived at a very good solution.
If you like, you may now push a changelog entry and a minor version and I'll merge and tag a release.

ulido:
Great, thanks @JonasIsensee ! I've merged the latest master in and updated the version number and changelog.

PR #288, opened 2021-09-10
Added support for returning `OrderedDict` when using `ntuple2dict`.

rs7q5:
Missed the addition of supporting `OrderedDict` in `ntuple2dict with pull request #283.

### DIFF
--- a/src/naming.jl
+++ b/src/naming.jl
@@ -346,7 +346,7 @@ end
 # Credit of `ntuple` macro goes to Sebastian Pfitzner, @pfitzseb
 
 """
-    ntuple2dict(nt) -> dict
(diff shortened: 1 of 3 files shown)


src/naming.jl
```diff
@@ -349,7 +349,8 @@ end
     ntuple2dict(nt) -> dict
```
Datseris, line 349: add in the docstring the `[DT = Dict,] nt` to make this visible for others?

rs7q5: I used `[type = Dict,]` to keep consistent with `struct2dict` docstring.

Datseris:
great thanks!

PR #290, opened 2021-09-10
use environment variables for tagging

### LINKED ISSUE #289: Global ENV variables for default values of tag, storepatch, etc.
I find myself putting `storepatch = false` in every call to saving functions from DrWatson (because I've had so many problems in the past with the patch storing functionality, and because I personally do not use it). 

At the moment there is no way to say globally to drwatson "use this default". However, using a couple of environment variables and putting e.g. `storepatch = get(ENV, "DRWATSON_STOREPATCH", true)` seems rather straightforward.

Datseris:
Hope I did everything right!

Closes #289

### DIFF
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -24,7 +24,8 @@ end
 
 ## Keywords
 * `suffix = "jld2", prefix = default_prefix(config)` : Used in [`savename`](@ref).
-* `tag::Bool = istaggable(suffix)` : Save the file using [`tagsave`](@ref) if `true`.
+* `tag::Bool = get(ENV, "DRWATSON_TAG", istaggable(suffix))` : Save the file 
+  using [`tagsave`](@ref) if `true` (which is the default).
 * `gitpath, storepatch` : Given to [`tagsave`](@ref) if `tag` is `true`.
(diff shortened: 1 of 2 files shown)


Datseris:
@JonasIsensee I've noticed that `collect_results` doesn't do any tagging when saving the dictionary. I guess it's on purpose? Should we add tagging there? 

Maybe it is not so helpful per se to know the state of the code for producing the results dictionary. What's more important is to know the state of the code when producing each result...? Then again, it doesn't cost much to just replace `wsave` with `tagsave` in 
```
!newfile && wsave(filename, Dict("df" => df))
```

JonasIsensee:
> @JonasIsensee I've noticed that `collect_results` doesn't do any tagging when saving the dictionary. I guess it's on purpose? Should we add tagging there?
> 
> Maybe it is not so helpful per se to know the state of the code for producing the results dictionary. What's more important is to know the state of the code when producing each result...? Then again, it doesn't cost much to just replace `wsave` with `tagsave` in
> 
> ```
> !newfile && wsave(filename, Dict("df" => df))
> ```

I suppose it was on purpose. My mental model of the `collect_results` dataframe was always, that it should be rather *state-less*. So, data deserves a tag but the analysis doesn't.

PR #293, opened 2021-10-01
Use `jldopen` in `collect_results!` when scanning JLD2 files.

ulido:
Employ lazy loading via Requires.jl to allow special load functions for particular file types in `collect_results!`, and falling back to `wload` as a default. This potentially makes scanning collections of large files much faster when using the `black_list` argument of `collect_results!`.

An example use case are simulations that generate amounts of data where the parameters are stored in a Dict, together with the data under the keyword `data`, all saved in a JLD2 file. If we only want to enumerate all the parameters that have been run, we would use `collect_results(datadir("simulations"), black_list = ("data",))`. With `wload`, this still loads the whole file into memory before ignoring the `data` key. With `jldopen`, only the parameters are loaded into memory. In my tests, this leads to around 400x speed up when scanning large collections of files.

For now, only `jldopen` for JLD2 files is implemented, but this could potentially be extended to other data storage formats that allow memory mapped access to storage keys.

### DIFF
--- a/src/DrWatson.jl
+++ b/src/DrWatson.jl
@@ -45,6 +45,19 @@ using Requires
 function __init__()
     @require DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0" begin
         include("result_collection.jl")
+        @require JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819" begin
+            using .JLD2
+            # This allows us to use the much faster jldopen in collect_results
+            function load_jld2(f, path)
+                @debug "Opening $(path) with jldopen."
+                jldopen(path) do data
+                    return f(data)
+                end
+            end
+            ext_register[".jld2"] = load_jld2
+            # JLD2 doesn't define keytype, but only supports Strings anyway.
+            Base.keytype(f::JLD2.JLDFile) = String
+        end
     end
 end
 
--- a/src/result_collection.jl
+++ b/src/result_collection.jl
@@ -148,8 +148,8 @@ function collect_results!(filename, folder;
         # Now update the mtime of the new or modified file
         mtimes[file] = mtime_file
 
-        data = rpath === nothing ? wload(file) : wload(joinpath(rpath, file))
-        df_new = to_data_row(data, file; kwargs...)
+        fpath = rpath === nothing ? file : joinpath(rpath, file)
+        df_new = to_data_row(fpath; kwargs...)
         #add filename
         df_new[!, :path] .= file
         if replace_entry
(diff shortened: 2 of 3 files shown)


src/DrWatson.jl
```diff
@@ -45,6 +45,19 @@ using Requires
...
+            function load_jld2(f, path)
+                @debug "Opening $(path) with jldopen."
+                jldopen(path) do data
+                    return f(data)
+                end
+            end
+            ext_register[".jld2"] = load_jld2
+            # JLD2 doesn't define keytype, but only supports Strings anyway.
+            Base.keytype(f::JLD2.JLDFile) = String
+        end
```
Datseris, line 60: Hm, something to keep in mind here. This code becomes available if a user has `using JLD2` in the start of their script. 

I, for example, do not do this. I don't `using JLD2`, even though I save with JLD2. As long as JLD2 is installed, FileIO does not require you to actually use the package. So perhaps we need to think of a different design that doesn't go this Requires route. Maybe we can have and `if` clause inside `collect_results`, so that if the file endings are jld2 it uses your faster method?

ulido: I did this intentionally such that JLD2 doesn't become a hard dependency. It would actually be better to first check if JLD2 is available and if yes, load it, but I do not know the canonical way for this in Julia (but I just need to find the time to check how FileIO does it).

Right now, if JLD2 isn't loaded, the code will fall back to `wload` for the first file. This triggers JLD2 to be loaded via FileIO. From the second file onward, the fast path via `jldopen` is available and will used.

Datseris: Huh, are you sure about that? Have you actually tested it that it works this way? I don't know for sure.

JonasIsensee:
Hi, thanks for your work!
I away currently but I'll try to find the time for a review later this week.

src/DrWatson.jl
ulido: Yes, I tested this in my workflow. But I didn't write a test for it because that would require a reorganisation of `tests/update_results_tests.jl`. I'd need to run that test before the `using JLD2` statement at the top.

What FileIO does is call `Base.require` inside a `try` block. The code is a bit obscure because it's all wrapped in macros, but I think this would also be a feasible route to take.

ulido:
> Hi, thanks for your work! I away currently but I'll try to find the time for a review later this week.

Ok, thanks, sounds good to me!

src/DrWatson.jl
Datseris: OOOooh okay, thanks for being thorough with this, one can never be certain haha! 

Okay, then I'll let Jonas review and merge this, given his expertise on the subject. All criticism I had to give I gave!

ulido:
In the latest commit I got rid of the code triggered by Requires.jl and used `Base.require` inside a `try` block to load JLD2 and set up the loading functions. This now has the advantage of automatically loading JLD2 if it is available. The only downside is that this now happens if and when `DataFrames` is loaded, meaning that if `DataFrames` is loaded before the project environment, JLD2 might not yet be available (depending on if the user has it installed globally). But loading packages before the DrWatson environment is activated is not recommended per the documentation anyway, so I consider that less of a problem.

src/DrWatson.jl
ulido: Great, thank you!

ulido:
In the latest commit, I added `JLD2` as a direct dependency (requiring version `0.4.15` which defines `keytype`). This massively simplified the code. Together with using `FileIO.query` (thanks for the suggestion @JonasIsensee) this now allows simple method dispatch to select the right `to_data_row` code path.

Let me know what you think, and if you are happy, I'll update version and changelog.

PR #299, opened 2021-10-21
Fix drwatson as author for project setup git commits

### LINKED ISSUE #298: Unexpected local .git config values lead to loss of authorship record
**Describe the bug**
When a project is initialized with DrWatson (git=true), `.git/config` will have values set under `[user]` with `name = DrWatson` and `mail = no@mail`.  As a result, all commits from this repository will be attributed to DrWatson. Most guidance has users set their git names globally (i.e. `git config --global user.name`), and this global setting will be unexpectedly overwritten by DrWatson. This can result in unrecoverable loss of authorship records if unnoticed.

**Minimal Working Example**
Initialize a DrWatson project with `git=true` and commit some change. `git log` will show `Author: DrWatson <no@mail>`

DrWatson v2.6.2
Julia v1.6.3

grahamas:
Closes #298

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.7.0"
+version = "2.7.1"
 
 [deps]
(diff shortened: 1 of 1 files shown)


grahamas:
I added tests that check what the global user.name and user.email are before `initialize_project` and then confirm that those are still effective from within the new repo.

PR #302, opened 2021-10-31
taking automation to 11 example

Datseris:
I've realized that the true power of `savename` and `produce_or_load` is not as commonly understood. Users are searching for some kind of external simulation helpers to do what is in principle already possible with these two functions. I've been using this workflow in many papers and it really does make things flow very smoothly. Perhaps this example can help people see the light and minimize their copy pasting!


Datseris:
this review has an update message as well, can someone please review it?

Datseris:
@tamasgal  or  @JonasIsensee   or @sebastianpech  or @SebastianM-C  could one of you review this please?

SebastianM-C:
Hi, I tried adding the package in a new environment, but I'm not seeing the update message
```
(@v1.6) pkg> activate --temp
  Activating new environment at `C:\Users\SEBAST~1\AppData\Local\Temp\jl_9gh9Ja\Project.toml`

(jl_9gh9Ja) pkg> add DrWatson#advcanced_docs
    Updating git-repo `https://github.com/JuliaDynamics/DrWatson.jl.git`
    Updating registry at `C:\Users\sebastian\.julia\registries\General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
    Updating registry at `C:\Users\sebastian\.julia\registries\JuliaSimRegistry`
   Resolving package versions...
    Updating `C:\Users\sebastian\AppData\Local\Temp\jl_9gh9Ja\Project.toml`
  [634d3b9d] + DrWatson v2.7.3 `https://github.com/JuliaDynamics/DrWatson.jl.git#advcanced_docs`
    Updating `C:\Users\sebastian\AppData\Local\Temp\jl_9gh9Ja\Manifest.toml`
  [34da2185] + Compat v3.40.0
  [864edb3b] + DataStructures v0.18.10
  [634d3b9d] + DrWatson v2.7.3 `https://github.com/JuliaDynamics/DrWatson.jl.git#advcanced_docs`
  [5789e2e9] + FileIO v1.11.2
  [033835bb] + JLD2 v0.4.15
  [1914dd2f] + MacroTools v0.5.9
  [bac558e1] + OrderedCollections v1.4.1
  [189a3867] + Reexport v1.2.2
  [ae029012] + Requires v1.1.3
  [6c6a2e73] + Scratch v1.1.0
  [3bb67fe8] + TranscodingStreams v0.9.6
  [3a884ed6] + UnPack v1.0.2
  [0dad84c5] + ArgTools
  [56f22d72] + Artifacts
  [2a0f44e3] + Base64
  [ade2ca70] + Dates
  [8bb1440f] + DelimitedFiles
  [8ba89e20] + Distributed
  [f43a241f] + Downloads
  [b77e0a4c] + InteractiveUtils
  [b27032c2] + LibCURL
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [37e2e46d] + LinearAlgebra
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [a63ad114] + Mmap
  [ca575930] + NetworkOptions
  [44cfe95a] + Pkg
  [de0858da] + Printf
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA
  [9e88b42a] + Serialization
  [1a1011a3] + SharedArrays
  [6462fe0b] + Sockets
  [2f01184e] + SparseArrays
  [10745b16] + Statistics
  [fa267f1f] + TOML
  [a4e569a6] + Tar
  [8dfed614] + Test
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
  [deac9b47] + LibCURL_jll
  [29816b5a] + LibSSH2_jll
  [c8ffd9c3] + MbedTLS_jll
  [14a3606d] + MozillaCACerts_jll
  [83775a58] + Zlib_jll
  [8e850ede] + nghttp2_jll
  [3f19e933] + p7zip_jll
Precompiling project...
  1 dependency successfully precompiled in 3 seconds (14 already precompiled)

julia> using DrWatson

julia>
```

I'm not sure if I understand when the message should be printed. I would expect that to happen during precompilation, or at least on the first `using`.

I gave https://docs.julialang.org/en/v1/manual/modules/#Module-initialization-and-precompilation another read and I tried using `--compiled-modules=no` and this time I got the update message on the first `using`. This happens both on the latest release and on this PR. What I understand from this is that the `update_message` code is called during precompilation and I'm not sure why, but with the parallel precompilation in julia 1.6+, the message doesn't show up.

Using the update message during precompilation might not be ideal in this case, but maybe using `__init__` could help with this, since it will happen at runtime after the first `using`. The Scratch.jl readme also indicates this approach of using `__init__`. I think something like the following should work, but I haven't tested it yet

```julia
function __init__()
    versions_dir = @get_scratch!("versions")

    if !isfile(joinpath(versions_dir, update_name))
        printstyled(stdout, update_message)
        touch(joinpath(versions_dir, update_name))
    end
end
```

Datseris:
Hi @SebastianM-C thanks for the feedback. Oddly, it has worked for me locally without the `init`. I've tried the `init` version, could you please try again?

SebastianM-C:
It works now. I also see it in CI: https://github.com/JuliaDynamics/DrWatson.jl/runs/4115369758?check_suite_focus=true#step:7:201
and checking the CI for the previous version (e.g. https://github.com/JuliaDynamics/DrWatson.jl/runs/4105322032?check_suite_focus=true#step:7:199), you can also see that the message did not show up.

The reason for it working on your machine might be that you are developing the package and you are using the serial precompilation from `using` and not the one from Pkg operations (like `]add` or `]up`).

Datseris:
yes, you are correct, thanks.

src/DrWatson.jl
```diff
@@ -43,31 +43,39 @@ include("dict_list.jl")
...
+        printstyled(stdout,
+        """
+        \nUpdate message: DrWatson v$update_version
+
+        * New section "Taking project input-output automation to 11" in the documentation.
+          It showcases how to eliminate code duplication and streamline your simulation setup
+          and run phase using `savename` and `produce_or_load`.
+        * By default now `gitpatch` is NOT saved when calling `tag!` and derivative functions.
+          This is due to an unknown problem that causes collecting the git patch to 
+          never hault, potentially not saving a user's output.
```
JonasIsensee, line 72:
```suggestion
          never halt, potentially not saving a user's output.
```

docs/src/real_world.md
```diff
@@ -514,3 +513,36 @@ julia> expensive_computation(5)
...
 2021-05-19 19:20:29 | n = 5 | error = 0.15 | maxrss = 326.27 MiB
 ```
+
+## Taking project input-output automation to 11
+The point of this section is to show how far one can take the interplay between [`savename`](@ref) and [`produce_or_load`](@ref) to **automate project input-to-output and eliminate as many duplicate lines of code as possible**. Read [Customizing `savename`](@ref) first, as knowledge of that section is used here.
+
+The key ingredient is that [`produce_or_load`](@ref) was made to work well with [`savename`](@ref). You can use this to automate the input-to-output pipeline of your project by following these steps:
+1. Define a custom struct that represents the input configuration for an experiment or a simulation. 
+2. Extend [`savename`](@ref) appropriately for it. 
+3. Define a "main" function that takes as input an instance of this configuration type, and returns the output of the experiment or simulation as dictionary (We're not changing here the "default" way to save files in Julia as `.jld2` files. To save files this way you need your data to be in a dictionary with `Symbol` as keys).
```
JonasIsensee, line 523:
```suggestion
3. Define a "main" function that takes as input an instance of this configuration type, and returns the output of the experiment or simulation as dictionary (We're not changing here the "default" way to save files in Julia as `.jld2` files. To save files this way you need your data to be in a dictionary with `String` as keys).
```

(code also correctly uses `@strdict`)

JonasIsensee (commented):
Hi @Datseris,
this is really nice!

JonasIsensee:
Ah, :grimacing:  now I was too slow

Datseris:
yeap can you do a quick push?

PR #303, opened 2021-11-11
Fix recursively_clear_path for directories

jonas-schulze:
Suppose you would like to save a custom data structure as a directory containing several files (or other directories). This fixes the `wsave` machinery for that use case.

I am working on a cluster and my result type holds `Future`s (non-cached, i.e. having `fut.v == nothing`, though I'm considering to move to [DeferredFutures.jl](https://github.com/invenia/DeferredFutures.jl)) to data on different nodes. I would like to have each node/worker write a separate file to disk, such that I don't need to transfer all the data to the managing process.

If there is anything I missed, or you've got any suggestions, I'm all ears. 🙂

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.7.3"
+version = "2.7.4"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -222,9 +222,9 @@ end
 
 #recursively move files to increased backup number
 function recursively_clear_path(cur_path)
-    isfile(cur_path) || return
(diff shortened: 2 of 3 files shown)


jonas-schulze:
I saw that `wload` is currently forwarded to `FileIO.load`. Maybe one could introduce an optional first argument designated for the return type, that allows a user to define custom methods for `wload`:

```julia
wload(args...; kwargs...) = FileIO.load(args...; kwargs...)
wload(::Type, args...; kwargs...) = FileIO.load(args...; kwargs...)

# custom types:
wload(::Type{Composite}, dir) = wload(joinpath(dir, "x.jld2"))
```

`produce_or_load` would then have an optional kwarg `type::DataType=Nothing` for the return type, which could be wired into `wsave` (if not `Nothing`, assert that `f` produces the correct type: `f()::type`). This should allow for such nested file structures.

For my particular use case, unfortunately, I don't see an easy way to get `produce_or_load` running, as it would have submit jobs to the cluster. I would rather like to avoid that.

Datseris (approved):
Easy fix. Please increment patch version in Project.toml!

Regarding your suggestion for `wload`: yes, I like it very much. But please do this in a separate PR!

PR #305, opened 2021-11-16
Fix ENV usage

jonas-schulze:
If `get(ENV, ...)` returns the non-default value, that will be a `String`. Therefore, it first has to be parsed as a `Bool`.

### DIFF
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -24,7 +24,7 @@ end
 
 ## Keywords
 * `suffix = "jld2", prefix = default_prefix(config)` : Used in [`savename`](@ref).
-* `tag::Bool = get(ENV, "DRWATSON_TAG", istaggable(suffix))` : Save the file
+* `tag::Bool = readenv("DRWATSON_TAG", istaggable(suffix))` : Save the file
   using [`tagsave`](@ref) if `true` (which is the default).
 * `gitpath, storepatch` : Given to [`tagsave`](@ref) if `tag` is `true`.
 * `force = false` : If `true` then don't check if file `s` exists and produce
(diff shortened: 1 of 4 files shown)


jonas-schulze:
Can this still be included in the `v2.7.4` release? 🙂

Datseris:
oh wow, pretty advance knowledge of `ENV` handling that I obviously not have :D Please increment the Project.toml patch version, and if possible add a test in the test suite!

jonas-schulze:
The functionality involving `ENV` haven't been tested at all before. I only added some tests for `add!` as that is what these flags are essentially handed down to. Technically, that doesn't test that `storepatch` etc are properly wired, but as their usage pattern (involving the new `readenv`) is identical, I hope that's fine.

Did you see https://github.com/JuliaDynamics/DrWatson.jl/pull/305#issuecomment-970510062? As far as I can see, you haven't released `v...4` yet, which I marked in my previous commit (#303). So I think it would be odd to mark this as `v...5` without having a `v...4` released. Therefore I didn't bump the patch version.

Datseris:
Just testing `tag!` is enough, yes, but tests fail at the moment.

> Did you see #305 (comment)? 

OOoooooooh sorry about that, I must have forgotten to tag the release. Okay, once we merge this i tag the .4 release!

jonas-schulze:
The tests are passing locally for me. It looks like CI doesn't expect `git` to perform a commit, so it's not completely configured. I might have found a solution. 🤞

src/saving_files.jl
```diff
@@ -24,7 +24,7 @@ end
 
 ## Keywords
 * `suffix = "jld2", prefix = default_prefix(config)` : Used in [`savename`](@ref).
-* `tag::Bool = get(ENV, "DRWATSON_TAG", istaggable(suffix))` : Save the file
+* `tag::Bool = readenv("DRWATSON_TAG", istaggable(suffix))` : Save the file
```
Datseris, line 27: I now realize the problem here. `readenv` is neither an exported name, nor does it have a docstring. How does it help a user to see `readenv` here? They know nothing about it, nor they can learn anything about it. I don't think this is useful for a documentation string. 

Not sure what the best way is, but I definitely prefer the original way. Even though it is not "correct", it does convey what is happening.

jonas-schulze: I agree that it's not the ideal solution right now. Though reading the docstring was a red herring to me, and that caused me to do this PR. I wasn't using this functionality but had been working with `ENV` in another project. Leaving it as it was before will lead potentially many other users to read the code and "try to fix this" even though it has already been fixed.

I would instead just add a docstring to `readenv` but I wouldn't export it.

Datseris: okay, let's add a docstring to `readenv`, and also replace it with `DrWatson.readenv` in the docs!

jonas-schulze: > also replace it with DrWatson.readenv in the docs!

Had the same idea. 😄

PR #310, opened 2022-01-02
Fix produce_and_load bug

### LINKED ISSUE #309: @produce_or_load sometimes doesn't produce a script field
**Describe the bug**
`@produce_or_load(f, path, config)` fails to add a `script` field to the `Dict` that is saved / output when the value-type of the dictionary returned by `f` is such `DrWatson.scripttag!` must widen the value-type in order to tag it.

**Minimal Working Example**
```julia
using DrWatson

function write_a_file(config)
    return Dict(
        "field" => config,
    )
end

@produce_or_load(write_a_file, "results", Dict("a" => 5.0))
```
yields
```julia
(Dict("field" => Dict("a" => 5.0)), "results/a=5.0.jld2")
```

This is on version 2.7.5 of `DrWatson`.

**Cause + Fix**

The cause of the problem is the interaction between here: https://github.com/JuliaDynamics/DrWatson.jl/blob/669be8894915d85adf814ffb8b6f737ffa2a83f9/src/saving_tools.jl#L256

and here: https://github.com/JuliaDynamics/DrWatson.jl/blob/f46c6318a0acd8570bd4562605358478c4e38d64/src/saving_files.jl#L116

Specifically, if the type of the value field of the dictionary needs to be widened in order to accept a `String`, a new `Dict` is necessarily allocated, and the `script` string added to this new `Dict`.

One possible fix is to replace
```julia
scripttag!(data, $s; gitpath = gitpath)
```
in the line linked above, with
```julia
data = scripttag!(data, $s; gitpath = gitpath)
```

I'm happy to open a PR with the fix if a maintainer could indicate the optimal location for a corresponding regression test.

edit: typo

willtebbutt:
Resolves #309

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.7.5"
+version = "2.7.6"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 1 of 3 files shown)


Datseris:
great, thanks, please also increase patch version in Project.toml!

Issue #311, opened 2022-01-04
Allow filtering file names with `collect_results`


junglegobs:
I would like to be able to filter the results loaded which don't match a certain file name, since loading all the files in my folder of interest is quite time consuming.

Consider the following 3 files in a folder:

```
a=1_b=1.jld2
a=2_b=1.jld2
a=3_b=2.jld2
```

I would like to be able to write `collect_results(dir; exclude=[r"b=2"])` to only load files whose names match that Regex expression. Similarly you could perhaps think of doing `collect_results(dir; include=[r"b=1"])`.

Given the smart naming of files that DrWatson provides, this doesn't seem to outlandish no?

Datseris:
Nope, it actually seems very sensible and rather straightforward to implement in code if you are willing to open a PR :)

junglegobs:
Done!

PR #312, opened 2022-01-05
Add filtering file names when collecting results

junglegobs:
See #311


src/result_collection.jl
```diff
@@ -84,6 +86,8 @@ function collect_results!(filename, folder;
     verbose = true,
     update = false,
     newfile = false, # keyword only for defining collect_results without !
+    rinclude = [r""],
+    rexclude = [r"^\b$"],
```
Datseris, line 90: What do the default regex expressions mean? I thought by default we excluded nothing...?

Datseris (commented):
Please also add a CHANGELOG entry and increment minor version in Project.toml.

Datseris:
@junglegobs It may be useful in this PR to add a documentation entry in the "real world examples". A simple section that says "I only want to collect simulations that contain b = 2" or so would be already a good entry.

test/update_results_tests.jl
```diff
@@ -64,6 +64,19 @@ cres_relpath = collect_results!(relpathname, folder;
     rpath = projectdir())
 @info all(startswith.(cres[!,"path"], "data"))
 
+###############################################################################
+#                           Exclude or exclude files                          #
```
Datseris, line 68:
```suggestion
#                           Include or exclude files                          #
```

src/result_collection.jl
```diff
@@ -116,6 +120,18 @@ function collect_results!(filename, folder;
     else
         allfiles = joinpath.(Ref(folder), readdir(folder))
     end
+    
+    # Filter through entries
+    idx_del = Int[]
+    for i in eachindex(allfiles)
+        file = allfiles[i]
+        include_bool = any(match.(rinclude, file) .!== nothing)
+        exclude_bool = any(match.(rexclude, file) .!== nothing)
```
Datseris, line 129: Can we do this a bit more performantly so we don't allocate an array at each call? any can be used with iterators as well. `any(x == 2 for x in y)`

src/result_collection.jl
junglegobs: That's what those 2 expressions do - Regex can be surprisingly convoluted.

On that note though, I will skip the filtering if the Regex matches the expressions above since it doesn't do anything.

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# 2.7.7
```
Datseris, line 1: Since this adds a new feature, it needs to increment the minor version, not the patch version (second number, not third). See also https://semver.org/

test/update_results_tests.jl
```diff
@@ -65,11 +65,13 @@ cres_relpath = collect_results!(relpathname, folder;
 @info all(startswith.(cres[!,"path"], "data"))
 
 ###############################################################################
-#                           Exclude or exclude files                          #
+#                           Ixclude or exclude files                          #
```
Datseris, line 68:
```suggestion
#                           Include or exclude files                          #
```

junglegobs:
Hey @Datseris I stopped following this for a while, but I think it should be good no?

Datseris:
I'm very sorry, I forgot to merge and tag...

Issue #315, opened 2022-01-10
Tooling for a make.jl file in the project root


goerz:
As explained in https://discourse.julialang.org/t/julia-based-makefile-replacement-for-research-workflows/74368, it might be nice to have some tooling that would allow me to write a script `make.jl` in the root of my research project that would automatically run the scripts to generate any missing output files or plots.

I realize this might a stretch, but it wouldn't seem entirely out-of-scope for a project like DrWatson to include something that takes a Dict defining a map of target files to dependencies and runs some code to produce those targets automatically.

Datseris:
This seems like it boils down to having a bunch of `.jl` files that use `produce_or_load`, and then a `make.jl` file what just does `incluide( all files )`. My question is, what do you do if for the _same_ `.jl` file you'd want to run this with _different_ parameters?

goerz:
I'm not sure we're exactly on the same page :-)

What I'm looking for is a way to automatically run the `.jl` files. Like a [`make.jl`](https://github.com/goerz-research/2022-01_Rydberg_Krotov_Spectral_Constraints/blob/master/make.jl) file that contains

~~~
RULES = [
    ScriptRule(
        [datadir("oct", savename("baseline_oct_result", Dict(:iter_stop=>100, :ν_max=>0.0), "jld2"))],
        scriptsdir("2022-01-09_baseline.jl")
    )
]
~~~

where the `2022-01-09_baseline.jl` script would be run *automatically* run if the jld2 doesn't exist. The script itself in this case still has `produce_or_load`, but that only takes care of not unnecessarily re-creating the `jld2` file if it already exists. It's somewhat redundant if someone is using the `make.jl`, but still important if they call the script directly.

The above snippet is currently in https://github.com/goerz-research/2022-01_Rydberg_Krotov_Spectral_Constraints, see also https://discourse.julialang.org/t/julia-based-makefile-replacement-for-research-workflows/74368/4?u=goerz

That system could actually specifically handle the situation of running the same `.jl` file with different arguments that you mention: You could have, e.g.,

~~~
RULES = [
    ScriptRule(
        [datadir("oct", savename("baseline_oct_result", Dict(:iter_stop=>100, :ν_max=>0.1), "jld2"))],
        scriptsdir("2022-01-09_baseline.jl"),
        args=[0.1]
    ), 
    ScriptRule(
[...]

PR #318, opened 2022-01-16
Refactor names of produce_or_load return values

goerz:
The current documentation

    produce_or_load([path="",] config, f; kwargs...) -> file, s

is quite confusing. It should be

    produce_or_load([path="",] config, f; kwargs...) -> data, filename

or

    produce_or_load([path="",] config, f; kwargs...) -> data, file

See https://github.com/JuliaDynamics/DrWatson.jl/issues/315#issuecomment-1013920021, https://github.com/JuliaDynamics/DrWatson.jl/issues/315#issuecomment-1013934772

In principle, this only affects the documentation, but it seems like a
good idea to refactor the names of the variables in the code to be
consistent.

### DIFF
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -1,14 +1,15 @@
 export produce_or_load, @produce_or_load, tagsave, @tagsave, safesave
 
 """
-    produce_or_load([path="",] config, f; kwargs...) -> file, s
-Let `s = joinpath(path, savename(prefix, config, suffix))` where `config` is some kind
-of named parameter container.
-If a file named `s` exists then load it and return it, along
-with the global path that it is saved at (`s`).
+    produce_or_load([path="",] config, f; kwargs...) -> data, filename
+Let `filename = joinpath(path, savename(prefix, config, suffix))` where
+`config` is some kind of named parameter container.
+If `filename` exists then load it and return the contained `data`, along
+with the global path that it is saved at (`filename`).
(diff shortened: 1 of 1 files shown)


goerz:
Right now, I'm doing

~~~
produce_or_load([path="",] config, f; kwargs...) -> data, filename
~~~

but there's actually quite a few places where `file` is used for the file name ([`tagsave`](https://juliadynamics.github.io/DrWatson.jl/dev/save/#DrWatson.tagsave), [`istaggable`](https://juliadynamics.github.io/DrWatson.jl/dev/save/#DrWatson.istaggable)). However, `filename` is also used in some places, e.g. [`safesave`](https://juliadynamics.github.io/DrWatson.jl/dev/save/#DrWatson.safesave).

I can easily change `filename` to `file` for `produce_or_load`. Maybe I should also commit something to make `filename`/`file` consistent in the other routines?

@Datseris: Any preference between `filename` or `file`?

Datseris:
Please use `filename`. If it affects only documentation, then this change is no problem. If there is both data and filename being mentioned, let's stick with what you suggest: `data, filename`.

goerz:
> Please use `filename`. If it affects only documentation, then this change is no problem. If there is both data and filename being mentioned, let's stick with what you suggest: `data, filename`.

Great, then this is should be ready for merging (or further review)

PR #319, opened 2022-01-17
Add dry_run option to produce_or_load

goerz:
See https://github.com/JuliaDynamics/DrWatson.jl/issues/315#issuecomment-1013954720

This is on top of #318, so I'll redo this PR once that is merged.

### DIFF
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -1,14 +1,15 @@
 export produce_or_load, @produce_or_load, tagsave, @tagsave, safesave
 
 """
-    produce_or_load([path="",] config, f; kwargs...) -> file, s
-Let `s = joinpath(path, savename(prefix, config, suffix))` where `config` is some kind
-of named parameter container.
-If a file named `s` exists then load it and return it, along
-with the global path that it is saved at (`s`).
+    produce_or_load([path="",] config, f; kwargs...) -> data, filename
+Let `filename = joinpath(path, savename(prefix, config, suffix))` where
+`config` is some kind of named parameter container.
+If `filename` exists then load it and return the contained `data`, along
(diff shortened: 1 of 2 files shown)


Datseris:
So the rationale is:

> The problem is that the output files generally depend on the arguments, so figuring out the output filenames requires running the code at some level. The only reasonable idea I've come up with would require produce_or_load/@produce_or_load to have a dry_run option, so that

> `produce_or_load([path="",] config, f; dry_run=true, kwargs...) -> exists, s`

> where exists is a boolean indicating whether the file already exists or not, and s is the full path to the output file, as before. With that, scripts could self-report their output file(s): They'd have to be run once in dry-run-mode when instantiating the rule, and then once more for the actual run.


Sorry, I missed that, because in #315 there is just so much text it is very difficult for me to go through every line... 

Unfortunately, I can't aggry with merging this in my right mind. You could just do `isfile(join(path, savename(config))` instead. which is literally a single line of code.

As I Said in https://github.com/JuliaDynamics/DrWatson.jl/issues/315#issuecomment-1014297437

> An overwhelming amount of options and keywords will swarm the ones that are truly useful, so we should be adding as much as possible only keywords that are genuinely useful.

and I don't think that is a good way to design software. We really shouldn't be adding keywords for something that can be so naturally done in a single line.

Datseris:
Don't take me closing this the wrong way, it just doesn't make sense to add this feature to `produce_or_load`, because, well, read the name of the function :D that's what its supposed to do, and now you say that with a different keyword the function should do something else entirely, and not only that, but that something else is trivially checked using the same arguments you give to the function anyways.

goerz:
That's okay, no hard feelings! :-)

I think I have some ideas on how to work around this...

PR #320, opened 2022-01-17
Add touch option to produce_or_load

### LINKED ISSUE #317: Touch outputfile in `produce_or_load`
Loosely related to #315

Suppose someone has a `Makefile` or similar system to (re-)run a script if the `jdl2` output file the script produces has an older modification time than the script. However, the script uses `@produce_or_load`, so it won't actually do anything to the file (just "load" it). Consequently, the next time `make` (or whatever) runs, it will still see the same modification times and run the script again.

It might be nice if `@produce_or_load` had an option `touch=true/false` so that it touches (as in the [`touch`](https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.touch) function) the output `jdl2` file in the "load" case.

This behavior might be considered debatable. After all, if the script changes significantly, there's no guarantee that it would still produce the same output, and touching the output file would silently repress that. However, I'm thinking mostly of refactoring changes here that do not change the output. For more substantial changes, I'd manually deleted the `jdl2` file, causing `make` to recreate it.

goerz:
Closes #317

This is based on top of #318 and #319, so this is just a preview until those are merged

### DIFF
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -1,14 +1,15 @@
 export produce_or_load, @produce_or_load, tagsave, @tagsave, safesave
 
 """
-    produce_or_load([path="",] config, f; kwargs...) -> file, s
-Let `s = joinpath(path, savename(prefix, config, suffix))` where `config` is some kind
-of named parameter container.
-If a file named `s` exists then load it and return it, along
-with the global path that it is saved at (`s`).
+    produce_or_load([path="",] config, f; kwargs...) -> data, filename
+Let `filename = joinpath(path, savename(prefix, config, suffix))` where
+`config` is some kind of named parameter container.
+If `filename` exists then load it and return the contained `data`, along
+with the global path that it is saved at (`filename`).
 
-If the file does not exist then call `file = f(config)`, with `f` your function
-that produces your data. Then save the `file` as `s` and then return `file, s`.
+If the file does not exist then call `data = f(config)`, with `f` your function
+that produces your data. Then save the `data` as `filename` and then return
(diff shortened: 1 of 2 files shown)


Datseris:
Can you please rebase this with `master`? And then I can merge? But also please 

1. add a changelog.md entry
2. increment minor version in Project.toml.

Datseris:
Actually, let's think this a bit more. The justification is to solve:

> script if the jdl2 output file the script produces has an older modification time than the script. However, the script uses @produce_or_load, so it won't actually do anything to the file (just "load" it). 

Okay, but, why should this matter? If the script was modified in a way that the data results are affected, then you should anyways use the keyword `force = true` when using `produce_or_load` (what I do for example). if the script is modified in a way that the output data aren't anyways affected, why would it matter if the data has "older save time" than the script?

Actually, you already say the problem:

> This behavior might be considered debatable. After all, if the script changes significantly, there's no guarantee that it would still produce the same output, and touching the output file would silently repress that.

So I'd vote to close this as well in favor of `force` keyword in case the data should be re-created.

---

> Consequently, the next time make (or whatever) runs, it will still see the same modification times and run the script again.

That seems to be a design problem with the `make` pipeline though. Why not just design the `make` pipeline so that it doesn't care about the dates? Caring about the dates doesn't make sense to begin with.

goerz:
That's fine, `touch` isn't really necessary. I can easily handle this in the make-code.

And yeah, looking at modification times doesn't make much sense in the context of a script that uses `produce_or_load`.

PR #323, opened 2022-01-20
Add filename option for produce_or_load

### LINKED ISSUE #322: Decouple `produce_or_load` and `savename`
## Rationale

There are two situations where it could be nice to circumvent the internal use of `savename` in `produce_or_load` and to directly pass it a full file name:

1. If the filename is already used in the script before the call to `produce_or_load` with an explicit call to `savename`, it can make the code slightly cleaner to pass that existing filename to `produce_or_load`. It would also eliminate the possibility of bugs due to any inconsistency between the two calls.

2. If for some reason a project does not want to interpolate the key-value pairs from `config` into the output filename, e.g., if there are two many parameters (see #316), it might be difficult to "fight" the internal `savename` to give the desired filename.

## Implementation

In https://github.com/JuliaDynamics/DrWatson.jl/issues/315#issuecomment-1016106343, I proposed implementing an alternative method for `produce_or_load`. However, on further reflection that was an overly complicated solution. The "problem" is only a single line of code in `produce_or_load`, 

https://github.com/JuliaDynamics/DrWatson.jl/blob/4a92f3e1fc9dd062b0f44123266037b490d76c94/src/saving_files.jl#L54

So it seems easiest to slightly tweak the routine, maybe add a parameter to modify just that line.

I would consider the following possibilities:

1.  Add a `filename=nothing` keyword argument to `produce_or_load`

    The above line would simply turn into

    ~~~
    if isnothing(filename)
        filename = joinpath(path, savename(prefix, c, suffix; kwargs...))
    end
    ~~~

    Thus, if `filename` is given as a string, it will serve as the output filename directly.

    Advantages:

    * Easy to understand and explicit
    * Matches current documentation on `filename` being the return value (including the full path)

    Problems:

    * Requires `path=""`
    * Possibly clashes with `suffix`: suffix should be set to the file extension of `filename`
    * Ignores `prefix` and `kwargs...`. If these are non-emtpy, a warning should be printed, maybe even an `error`.

2. Allow `path` to include the filename directly. Whether `path` is a folder or a filename is determined by whether it has an extension.

    ~~~
    if splitext(path) == ""
        filename = joinpath(path, savename(prefix, c, suffix; kwargs...))
    else
        filename = path
    end
    ~~~

    Advantages:

    * Does not need any new parameter for `produce_or_load`
    * Very concise for the user

    Problems:

    * The behavior of `path` is rather subtle. Users might find this kind of implicit behavior confusing.
    * Possibly clashes with `suffix`: suffix should be set to the file extension of `path`
    * Ignores `prefix` and `kwargs...`. If these are non-emtpy, a warning should be printed.

3. Add a new `auto_filename=true` (or some other name, maybe `use_savename`, not sure). Use `path` as the full filename if `auto_filename=false`

    ~~~
    if auto_filename
        filename = joinpath(path, savename(prefix, c, suffix; kwargs...))
    else
        filename = path
    end
    ~~~

    Advantages:

    * Slightly more explicit than option 2

    Problems:

    * Still a little confusing for `path` to change meaning depending on `auto_filename`
    * Possibly clashes with `suffix`: suffix should be set to the file extension of `path`
    * Ignores `prefix` and `kwargs...`. If these are non-emtpy, a warning should be printed.

The above are in order of my preference (number 1 being my most preferred solution).

goerz:
Closes #322

### DIFF
--- a/src/saving_files.jl
+++ b/src/saving_files.jl
@@ -2,7 +2,8 @@ export produce_or_load, @produce_or_load, tagsave, @tagsave, safesave
 
 """
     produce_or_load([path="",] config, f; kwargs...) -> data, filename
-Let `filename = joinpath(path, savename(prefix, config, suffix))` where
+If not given as a keyword argument, let
+`filename = joinpath(path, savename(prefix, config, suffix))` where
 `config` is some kind of named parameter container.
 If `filename` exists then load it and return the contained `data`, along
 with the global path that it is saved at (`filename`).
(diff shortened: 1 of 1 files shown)


goerz:
Tests pending

PR #324, opened 2022-01-20
Add filename option for produce_or_load

goerz:
As discussed in https://github.com/JuliaDynamics/DrWatson.jl/issues/322#issuecomment-1017554130

This is a minimal implementation that adds a `filename` option that defaults to `savename(prefix, config, suffix, kwargs...)` and nothing else, as proposed by @Datseris 

It replaces the #323 PR

Because Julia doesn't support using the `kwargs...` in the default initializer for `filename` in the function signature, I had to use `filename=nothing`, and then initialize it in the body of the function. The spirit (and the documentation) is the same, though.

I've bumped the version to 2.9, since this seems like a "new feature", not just a bugfix


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# 2.9.0
+* Add `filename` option in `produce_or_load`, overriding the default filename generated by `savename`
```
Datseris, line 2:
```suggestion
* Add `filename` option in `produce_or_load`, possibly overriding the default filename generated by `savename`
```
The way it was phrased before it may have been misinterpreted as that we are actually changing the default behavior.

src/saving_files.jl
```diff
@@ -24,6 +24,8 @@ end
 ```
 
 ## Keywords
+* `filename = savename(prefix, config, suffix; kwargs...)` : Name of the file
+  to produce or load, relative to `path`. The output `filename` includes `path`.
```
Datseris, line 28: It is not necessary to say "The output `filename` includes `path`.", it is already stated in the previous part of the docstring. However, it would be useful to mention the reason for this keyword, i.e., your use case. Something like:

"This may be useful in situations where `config` has too many parameters for `savename` to be useful, and an explicitly specified name is more suitable."

src/saving_files.jl
```diff
@@ -24,6 +24,8 @@ end
 ```
 
 ## Keywords
+* `filename = savename(prefix, config, suffix; kwargs...)` : Name of the file
```
Datseris, line 27: It is problematic that this keyword is `filename` but also `filename` is the one with `joinpath(path, filename)`. We need to either state `filename` with a different name in the beginning of the documentation string, or use a different name for the keyword. 

Kind of off-topic, but I'd strongly recommend you to not include paths in `filename` and use the `path` argument explicitly. I used to do the same thing as you when we started developing DrWatson, but the other lead devs presented very straightforward arguments why one shouldn't do this. I don't remember these arguments anymore, they were something about "it doesn't work for all OS" or something. In any case, it isn't good practice.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.8.0"
+version = "2.9.0"
```
Datseris, line 4: Yes, this is perfectly correct, this is definitely a new feature.

CHANGELOG.md
goerz: Sounds good!

src/saving_files.jl
goerz: Sure, sounds good! I'll commit that.

What workflow do you prefer? Add a third commit with fixes based on the review, or amend my previous commits and force-push? Do you want me to squash the actual patch and the version/changelog update into a single commit?

src/saving_files.jl
goerz: Yeah, I actually had the same concern about `filename`... It's confusing to have it in two places with a slightly different meaning.

In https://github.com/JuliaDynamics/DrWatson.jl/pull/318#issuecomment-1013953382, we discussed whether the return value should be called `filename`, or just `file`. Maybe we should revisit that, and change it to `file`? That would mean we can use

~~~
file = joinpath(path, filename)
~~~

And, in the documentation, we would write

~~~
produce_or_load([path="",] config, f; kwargs...) -> data, file
~~~

(where `kwargs` includes `filename`), which would be much cleaner.

I also agree with you that keeping `filename` and `path` separate is generally a good idea.

Datseris: okay, let's go with `filename` for the name without the path and `file` for the joinpath-ed version.

src/saving_files.jl
Datseris: It does not matter, feel free to use whatever workflow suits you best. No matter what, I will squash-merge the pull request anyways, so all will be pushed into master as 1 commit.

goerz:
Ok, changed everything as we discussed. Should be ready for merge or another review

Datseris:
looks great, thanks!

goerz:
Thank you!

Sorry things got a little off-track in the original discussion. I'm glad we got to work it out... I'm very happy with the final result.

Incidentally, a completely unrelated question, if you don't mind. For my own Julia packages, I'm still figuring out the release process via JuliaRegistrator (I'm assuming you're using JuliaRegistrator as well). It seems that for DrWatson, you (automatically?) make a release for every merged PR. How do you do that? Is it somehow integrated with your CI? Do you manually tag JuliaRegistrator in the merge commit? Do you have a long-running "Release Issue", like what [I've been playing around with](https://github.com/JuliaQuantumControl/QuantumControl.jl/issues/3)?

Datseris:
I got to https://juliahub.com/ui/Home  , click Contribute > Register packages and give the link to the github repo. Unfortunately, I have not found a way to automate this further, nor would I like to :D I'd prefer to be the one deciding when a release should be submitted in the registry. You should also add TagBot and CompatHelper. Go to DrWatson/.github and copy all files there.

PR #329, opened 2022-01-21
Fix docs

Datseris:
Hopefully this will make docs buildable again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -428,3 +428,4 @@ test project
 juliadynamics-light.scss
 juliadynamics-dark.scss
 *.scss
+test_project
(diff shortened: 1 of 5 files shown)


Datseris:
well, that's odd as fuck. For some reason `rm(path, recursive = true, force = true)` does not actually delete the project folder set up by `initialize_project` any more. Well, here we go digging to hell why the hell this happens.

Datseris:
well the `rm` function is called, so why the hell is the directory remaining.

Datseris:
I don't understand whats going on. This project folder that wasn't removed by `rm` wasn't the problem. Something else is. The deployment complains that "warning: adding embedded git repository: previews/PR329/DrWatson Example
hint: You've added another git repository inside your current repository."

https://github.com/JuliaDynamics/DrWatson.jl/runs/4902252654?check_suite_focus=true#step:6:294

No real idea what this means.

Datseris:
Oh wait, DrWatson Example is generated in the `workflow.jl` tutorial.

PR #333, opened 2022-01-30
Add DrWatsonSim functionality

sebastianpech:
Next up

- [x] write docstrings
- [x] add examples from readme to DrWatson doc
- [ ] Include docstrings in the doc
- [ ] add simple "real world" example
- [ ] add "real world" example where a parameter config is loaded and altered and then started as a new simulation.


Datseris:
Alrighty, after the meeting with Alex, here are the actionable points:

1. A restrictive format of metadata storage has some important advantages: portability, security, standardization. 
2. This restrictive format would limit to the following types: `Int, Float64, Bool, String, Symbol`. And then, composite Julia types, as well as `Vector, Dict` of the basic types are also possible.
3. The output format for the matadata can be a Yaml format. There are several advantages for that: human readable, there is a Julia package that directly transforms dictionaries to Yaml, it is faster to search, and integrates with CaosDB.
4. Allowing the possibility for "additional fields" for each metadata field is important. Such fields could be a "string 

---

Other points of discussion:

Datseris:
@sebastianpech that's the write down, let me know what you think.

sebastianpech:
@Datseris I have not forgotten about this, just really focusing on finishing my phd at the moment. I'll post my thoughts and ideas about this in the next days.

Datseris:
@sebastianpech How is your PhD progressing? Have you defended? Any chance to get back to this?

sebastianpech:
@Datseris I defended my thesis, all done :) I will try to get back to it soon

Datseris:
Congratulations man, well deserved!!!

Datseris:
Bump here!

Issue #327, opened 2022-01-21
dict_list changes Dict type


msmerlak:
A dict of type, say, `{Symbol, Any}` such as `Dict{Symbol, Any}(:a => collect(1:10), :b => true)` will be turned into a `Vector{Dict{Symbol, Integer}}`. This can create unwanted behavior.

I propose the following modification to `_dict_list`:

```
function _dict_list(c::AbstractDict)

    dict = typeof(c)

    iterable_fields = filter(k -> typeof(c[k]) <: Vector, keys(c))
    non_iterables = setdiff(keys(c), iterable_fields)

    iterable_dict = dict(iterable_fields .=> getindex.(Ref(c), iterable_fields))
    non_iterable_dict = dict(non_iterables .=> getindex.(Ref(c), non_iterables))

    vec(
        map(Iterators.product(values(iterable_dict)...)) do vals
            dd = [k=>convert(eltype(c[k]),v) for (k,v) in zip(keys(iterable_dict),vals)]
            if isempty(non_iterable_dict)
                Dict(dd)
            elseif isempty(iterable_dict)
                non_iterable_dict
            else
                # We can't use merge here because it promotes types.
                # The uniqueness of the dictionary keys is guaranteed.
                dict(dd..., collect(non_iterable_dict)...)
            end
        end
    )
end
```

sebastianpech:
Yes, sounds reasonable to keep the original type. Could you open a PR?

Datseris:
yEAH I agree this is a bug!

PR #342, opened 2022-04-27
dict_list changed type, fixed

msmerlak:
Fixing issue #327

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "julia.environmentPath": "/Users/smerlak/Dropbox/github/DrWatson.jl"
+}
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "julia.environmentPath": "/Users/smerlak/Dropbox/github/DrWatson.jl"
+}
```
Datseris, line 3: I don't think this file should be committed here.

Datseris (changes requested):
Thanks a lot! It looks promising, however there are some things to be done.

- A test needs to be added in the test suite, otherwise how can we know the PR works :) Simply put as a test the very thing you reported in the issue.
- Incremenet patch number (last number) in the version in Project.toml
- Remove the settings.json file.

Datseris:
Bump here?

Issue #341, opened 2022-04-27
Provenance information for Plots.jl/PGFPlotsX.jl generated figures


bringuiers:
I'm relatively new to [DrWatson.jl](https://github.com/JuliaDynamics/DrWatson.jl) and started using it for a project. I have several scripts  that generate `pgf/tikz` plots saved in a `tikz` file. This is done using [Plots.jl](https://github.com/JuliaPlots/Plots.jl) with the [PGFPlotsX.jl](https://github.com/KristofferC/PGFPlotsX.jl) backend. The `tikz` files are then used in a LaTeX document within the same DrWatson.jl project (i.e. in papers) and if I regenerate the `tikz` files the names are the same. 

The `tikz` files are tracked via git however there are many instances where I want to view the script and list of data that was used to generate the `tikz` file. Rather than searching the git tree to find this out, I was thinking of injecting within the `tikz` file LaTeX comments that contain the git information but also a replica of the script and a list of data files in order to facilitate quick inspection. In the past I've done this by running shell commands within a Julia script to inject git info and the script content, but my question is to whether or not there is a more 'DrWatsonian` way to achieve my goal or maybe I'm missing something trivial that I need to do.

Datseris:
hi there, I am completely unfamiliar with tikz, but I'll try to help.

So, if I understood you correctly, you would like a way to use the `tag!` function for the `tikz` file, right? You can't use `tag!` directly, because it needs a dictionary, but wouldn't using https://juliadynamics.github.io/DrWatson.jl/dev/save/#DrWatson.@tag!  on an empty dict, and then simply extracting the dict's fields into the latex file do the trick? The same dict can contain the parameters you used. So, assuming that you have a julia file that generates the tikz plot, then I suggest:

```julia
d # dictionary with parameters for tikz plot
@tag!(d)

f = save_tikz_latex_file

fpr (key, val) in d
write_in_latex!(f, "key $(key) with val $(val)")
end
```

Or I didn't get what you want in the end?

Datseris:
By the way, if you solve this request, please consider adding a section about it in our "Real World Usage" example page :)

bringuiers:
Thanks for the quick reply. So I think,

> [@tag!(d)](https://juliadynamics.github.io/DrWatson.jl/dev/save/#DrWatson.@tagsave)

Is very close to what I want! Now I just need to put together my preferred way to implement

> write_in_latex!(f, "key $(key) with val $(val)")

[...]

PR #343, opened 2022-05-06
Addition of saving plots to real-world doc portion.

bringuiers:
This is in response to #341, which adds a bit more context and detail.

### DIFF
--- a/docs/src/real_world.md
+++ b/docs/src/real_world.md
@@ -529,6 +529,92 @@ julia> expensive_computation(5)
 2021-05-19 19:20:29 | n = 5 | error = 0.15 | maxrss = 326.27 MiB
 ```
 
+## Saving plots
+At this stage after you have generated data and used [`collect_results`](@ref) to gather your data. Now you may be starting to make plots/figures and wanting to save them, then you can just overload the save tool `wsave` as discussed in [Saving Tools](@ref) to handle your plot data types. Most likely you will only get [`safesave`](@ref) automatically. Here is an example:
+
+!!! note
+    The example below assumes your using the [Plots.jl](https://github.com/JuliaPlots/Plots.jl) package and for the most part should work with any backend. The `Plots.savefig` function handles the image type based on the extension used, for example, `savefig(example,"example.png")`
+
+```julia
+using DrWatson
+using Plots
+DrWatson._wsave(filename,plot::T) where T<: Plots.Plot = begin
+    mkpath(dirname(filename));
+    savefig(plot,filename)
+end
+```
+The `filename` can be provided by [`savename`](@ref) to assist with adding parameters used to generate the data returned by [`collect_results`](@ref). Nothing new really here, but this raises the question, what if we want to get information from [`tagsave`](@ref) which provides critical details from `git` and the macro [`@tagsave`](@ref) gives us information about the script that ran and line number. We need to then work with text based plots.
+
+### Adding provenance details to text based plots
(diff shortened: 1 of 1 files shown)


docs/src/real_world.md
```diff
@@ -529,6 +529,92 @@ julia> expensive_computation(5)
...
+            outstr = "$k => $v"
+            insert_latex_comment(filename,outstr)
+        end
+    end
+end
+
+DrWatson.safesave(file, plot::T, info::Dict; kwargs...) where T<:Plots.Plot = begin
+    recursively_clear_path(file)
+    wsave(file,plot,info;kwargs...)
+end
```
Datseris, line 585: Wait, the point is that you are supposed to get `safesave` for free. Why is it that you need to code this? Is it because the original `safesave` method doesn't have `args...`? If so, just add `args...` to original function and we're good?

docs/src/real_world.md
```diff
@@ -529,6 +529,92 @@ julia> expensive_computation(5)
...
+
+DrWatson.tagsave(file, plot::T;
+                 gitpath=projectdir(),
+                 safe::Bool=DrWatson.readenv("DRWATSON_SAFESAVE", false),
+                 storepatch::Bool=DrWatson.readenv("DRWATSON_STOREPATCH", false),
+                 force=false, source=nothing, kwargs...
+                 ) where T<: Plots.Plot  = begin
```
Datseris, line 592: Ah yes, this is unfortunate but here we have to extend as we need an additional dictionary...

docs/src/real_world.md
```diff
@@ -529,6 +529,92 @@ julia> expensive_computation(5)
...
+end
+```
+The `filename` can be provided by [`savename`](@ref) to assist with adding parameters used to generate the data returned by [`collect_results`](@ref). Nothing new really here, but this raises the question, what if we want to get information from [`tagsave`](@ref) which provides critical details from `git` and the macro [`@tagsave`](@ref) gives us information about the script that ran and line number. We need to then work with text based plots.
+
+### Adding provenance details to text based plots
+ There are many instances where you will want to know what script generated a plot. This is hard to do consistently with plots stored in formats like `pdf` or `png`, but text-based vector formats like [`PGF/tikz`](https://en.wikipedia.org/wiki/PGF/TikZ) (or [`svg`](https://en.wikipedia.org/wiki/Scalable_Vector_Graphics)) are just mark-up text and allow for in-file comments. Therefore, one can store comments in the actual plot file providing provenance information which enables reproducing plots/figures more easily. To do this in `DrWatson`, we have to create a new [`tagsave`](@ref) method for our data type and `Dict` of `info` we want to write. Below is an example:
```
asinghvi17, line 549: I recently had the need to do a similar thing for PDF files, so if there's any interest in including this, the function is actually not too hard with `Ghostscript_jll`:

```julia
using Ghostscript_jll
const AUTHOR = Ref{String}("Anshul Singhvi")

function insert_pdf_metadata(
    file, desc::String;
    title = splitext(splitdir(path)[2])[1],
    author = AUTHOR[],
    keywords = [],
    additional_gs_flags = String[]
    )

    path = abspath(file)
    @assert endswith(path, ".pdf")
    @assert isfile(path)

    gs_options = Cmd(vcat(additional_gs_flags, [
        "-dBATCH",                      # run ghostscript in batch mode
        "-dNOPAUSE",                    # run without pause, exit immediately after complete
        "-dQUIET",                      # no output to stdout
        "-dProcessDSCComments=false",   # try to maintain metadata (this fails) see https://bugs.ghostscript.com/show_bug.cgi?id=693400
        "-sDEVICE=pdfwrite",            # use the `pdfwrite` device
        "-sOutputFile=-",               # output to stdout
    ]))

    gs_cmds = ``` $gs_options -f $path -c "[ /Subject ($desc) /Keywords ($(join(keywords, ", "))) /DOCINFO pdfmark"```

    pdf = Ghostscript_jll.gs() do gs_exe
        read(`$gs_exe $gs_cmds`, String)
    end

    write(path, pdf)

    return true

end
```

You can invoke the function like so:
```julia
path = "/path/to/pdf.pdf"
using Dates # just for the example, not required
insert_pdf_metadata(path, "This was written by Ghostscript!"; keywords = ["a = 1", "b = 1234.8403", "run 1 $(Dates.today())"])
```
The metadata displayed is:
Screen Shot 2022-05-07 at 10 22 03 AM
Screen Shot 2022-05-07 at 10 22 06 AM

Datseris:
sad :(

PR #352, opened 2022-07-17
Allow tests, docs, and CI hosting for projects

### LINKED ISSUE #130: Testing for scientifc projects
As the conversation in https://github.com/JuliaDynamics/DrWatson.jl/issues/89 was rather emotional I open a new issue for this.

I get, and agree, that the output of PkgTemplates.jl is too much for DrWatson. But shouldn't a scientific project have unit tests for the general functionality in `src/` as well? IMHO this is not related to packaging things up but just a good programming practice.

Looking forward to hear your opinion.

Datseris:
I am giving a workshop on Good Scientific Code this week. There, we are also showing students how to test their science projects, and how to document them. The need for a template for a **scientific project** (and NOT a "julia or python package") that can be tested and documentation-hosted automatically via CI became apparent. This PR achieves this for our `initialize_project`. Now, the project that is created can have a test folder, a docs folder, and a GitHub CI file. Importantly, the GitHub CI file does NOT expect the "standard Julia project structure", i.e., the project still doesn't have a UUID, nor a `src/ProjectName.jl` file, nor a `module` definition anywhere.

Notice that `]test ProjectName` still doesn't work, but this is a limitation of Julia internals, not our design or code, see https://github.com/JuliaDynamics/DrWatson.jl/issues/67 

Furthermore, over the years, I have re-thought #130 and I lean more and more into the importance of testing code bases. In my eyes, the best way to write science code is to put any generic function in an existing or a new package. However, sometimes it simply is not worth the effort. Nevertheless, all source code should be tested, or at least a motivation for testing should be there.

Closes #130


Datseris:
cc @carstenbauer as well.

Datseris:
Here's how the default project looks like: https://github.com/JuliaDynamics/ScienceProjectTemplate

Datseris:
Ping @tamasgal @sebastianpech or @carstenbauer if anyone can review this I would very much appreciate it :)

PR #355, opened 2022-08-15
Add ENV VAR check for displaying update msg

### LINKED ISSUE #354: Add ENV to suppress update messages
Using Dr. Watson in CI to auto generate documents with things like Quarto or Weave.jl leads to the update message being embedded in the resulting document.

Check OS ENV VAR to suppress printing the update message.

Remove update message all together 🤣

agerlach:
Closes #354 

I'm not quite sure how to actually test this. Also where should this be documented?


agerlach:
@Datseris Per our Slack discussion

src/DrWatson.jl
```diff
@@ -64,7 +65,7 @@ function __init__()
     @require DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0" begin
         include("result_collection.jl")
     end
-    if display_update
+    if env_var in keys(ENV) ? parse(Bool, ENV[env_var]) : display_update
```
Datseris, line 68: I think here we should define a proper block:
```julia
if env_var in keys(ENV)
    x = try 
        parse(Bool, ENV[env_var]) 
    catch
        display_update
    end
else 
    display_update
```
and use that output for safety

docs/src/project.md
```diff
@@ -137,3 +137,17 @@ pkg> activate .
 pkg> add Package1 Package2 ...
 ```
 Julia will automatically make the Project.toml and Manifest.toml files for you as you add packages used by your project.
+
+## Suppressing Update Messages
```
Datseris, line 141: Yeah, I am also not sure where to put this information. In the docs, especially here, it may be hard to discover. I was thinking to put this in the same bubble we have at the very start of `index.md` that shows the latest new features. But it is too long to put it there maybe...?

An alternative is to have this kind of information at the end of the update message itself (and I'll make sure I always retain this information there). However, let's make it as concise as possible. Let's say something as short as:
"pro hint: set the evironment variable `DRWATSON_UPDATE_MSG=0` (prior to `using DrWatson`) to turn off these update messages!"

I am not sure, let me know what you think.

src/DrWatson.jl
agerlach: Are OK w/ me just putting this logic in the `const display_update = true` line, i.e.

```julia
const display_update = if evn_ver in ...
   ...
else
   true
end

File: docs/src/project.md (hunk 2 above)
Comment (agerlach) [reply docs/src/project.md:141]: How about putting it somewhere in the docs with a linkable heading and then putting something in the update message like "To disable future update messages see http://..."

Comment (agerlach) [reply docs/src/project.md:141]: Seems fitting for a `Misc` or `FAQ` section if you had one.

File: src/DrWatson.jl (hunk 1 above)
Comment (agerlach) [reply src/DrWatson.jl:68]: Are OK w/ me just putting this logic in the `const display_update = true` line, i.e.

```julia
const display_update = if evn_ver in ...
   ...
else
   true
end

File: docs/src/index.md [hunk 3]
```diff
@@ -18,12 +18,17 @@ You can also watch this 8-minutes video that introduces DrWatson in JuliaCon2020
...
-To install, simply type `] add DrWatson` in your Julia session.
 DrWatson is part of [JuliaDynamics](https://juliadynamics.github.io/JuliaDynamics/), check out our [website](https://juliadynamics.github.io/JuliaDynamics/) for more cool stuff!
 
 !!! note "Star us on GitHub!"
     If you like DrWatson the please consider starring the [GitHub repository](https://github.com/JuliaDynamics/DrWatson.jl). This gives as an accurate lower bound of the number of people the software has helped!
 
+## Installing and Updating
+To install or update DrWatson, simply type `] add DrWatson` or `] up DrWatson`, respectively, in your Julia session.
+
+!!! note Pro-Tip
```
Datseris, line 29:
```suggestion
!!! note "Update messages"
```

docs/src/index.md
```diff
@@ -18,12 +18,17 @@ You can also watch this 8-minutes video that introduces DrWatson in JuliaCon2020
...
 DrWatson is part of [JuliaDynamics](https://juliadynamics.github.io/JuliaDynamics/), check out our [website](https://juliadynamics.github.io/JuliaDynamics/) for more cool stuff!
 
 !!! note "Star us on GitHub!"
     If you like DrWatson the please consider starring the [GitHub repository](https://github.com/JuliaDynamics/DrWatson.jl). This gives as an accurate lower bound of the number of people the software has helped!
 
+## Installing and Updating
+To install or update DrWatson, simply type `] add DrWatson` or `] up DrWatson`, respectively, in your Julia session.
+
+!!! note "Update messages"
+    Upon update DrWatson will display a message highlighting any major changes. To disable these messages set the environment variable `DRWATSON_UPDATE_MSG=0` prior to starting your Julia session.
```
Datseris, line 30: @agerlach is this true? Isn't it enough to set `DRWATSON_UPDATE_MSG=0` prior to `using DrWatson`? I could be after Julia starts. Right?

docs/src/project.md
Datseris: A FAQ section was a nice idea, but the way you did it now is also nice for me, having a section aobut install and update :) thanks a lot!

docs/src/index.md
agerlach: Julia reads the OS env vars and populates the `ENV` dict at startup. You could set `ENV["DRWATSON_UPDATE_MSG"]=0` after starting Julia and before `using DrWatson` though.

From https://docs.julialang.org/en/v1/manual/environment-variables/

> In Bash, environment variables can either be set manually by running, e.g., export JULIA_NUM_THREADS=4 before starting Julia

> Supposing that you want to set the environment variable JULIA_EDITOR to vim, you can type ENV["JULIA_EDITOR"] = "vim" (for instance, in the REPL) to make this change on a case by case basis, or add the same to the user configuration file ~/.julia/config/startup.jl in the user's home directory to have a permanent effect.

Datseris, line 30:
```suggestion
    Upon update DrWatson will display a message highlighting any major changes. To disable these messages set the environment variable `DRWATSON_UPDATE_MSG` to `0` prior to `using DrWatson`. 
```

Datseris:
great, thanks!

agerlach:
My pleasure!

PR #356, opened 2022-08-23
Make safesave use the full signature of wsave

daharn:
If using JLD2, this allows to use the name - value signature of fileio_save.

Since wsave already does this, the lack of ... here is probably just an oversight.


Datseris:
thanks so much, this was indeed an oversight!

Issue #358, opened 2022-09-27
Feature request: add macro @produced for simulation parameters that depend on other parameters.


aramirezreyes:
**Is your feature request related to a problem? Please describe.**
This feature request concerns the creation of dictionaries of parameters. From the documentation example:
```julia
using DrWatson
general_args = Dict(
    "model" => "barkley",
    "noise" => [0.075, 0.050, 0.025],
    "noisy_training" => true,
    "N" => 100,
)
dicts = dict_list(general_args)
```
Let's say we want to have another parameter, noise2, that for each simulation is double that of "noise" apart from another value.

Currently I would need to do:

```julia
map(dicts) do dict
    dict["noise2"] = 2*dict["noise"]
end
```

But this has the disadvantage of:

1. more verbose and
2.  the original `general_args` is now not enough to reproduce all the simulations.
3. It does not work if the we also want to have another value in all the simulations (see example below).

**Describe the solution you'd like**
It would be nice to have a `@produced` (name suggested by @Datseris) that allows to do something like:

```julia
using DrWatson
general_args = Dict(
    "model" => "barkley",
    "noise" => [0.075, 0.050, 0.025],
    "noise2" => [1.0, @produced("noise" -> 2*"noise")]
    "noisy_training" => true,
    "N" => 100,
)
dicts = dict_list(general_args)
```

This should create 6 different Dicts of simulation parameters:
```julia
julia> dicts = dict_list(general_args)
6-element Vector{Dict{String, Any}}:
 Dict("noise2" => 1.0, "model" => "barkley", "N" => 100, "noise" => 0.075, "noisy_training" => true)
 Dict("noise2" => 0.15, "model" => "barkley", "N" => 100, "noise" => 0.075, "noisy_training" => true)
 Dict("noise2" => 1.0, "model" => "barkley", "N" => 100, "noise" => 0.05, "noisy_training" => true)
 Dict("noise2" => 0.1, "model" => "barkley", "N" => 100, "noise" => 0.05, "noisy_training" => true)
 Dict("noise2" => 1.0, "model" => "barkley", "N" => 100, "noise" => 0.025, "noisy_training" => true)
 Dict("noise2" => 0.05, "model" => "barkley", "N" => 100, "noise" => 0.025, "noisy_training" => true)
 ```

**Describe alternatives you've considered**

Another option could be using `@onlyif` in the following fashion:
```julia
using DrWatson
general_args = Dict(
    "model" => "barkley",
[...]

PR #359, opened 2022-09-28
Adding Derived type for parameter dependency

aramirezreyes:
In this branch I will give it a go to implemented the feature described in #358. I have never programmed a macro so I will try to follow #169.  Advice from @sebastianpech is welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.10.1
+* Add `Derived` type and function to allow `dict_list` to compute some parameters from the value of other parameters.
+* Moved `export` statements to the top of the file in `dict_list.jl`
 # 2.10.0
 * Add `DRWATSON_UPDATE_MSG` ENV VAR for suppressing update messages
 # 2.9.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # 2.11.0
 * Add `Derived` type and function to allow `dict_list` to compute some parameters from the value of other parameters.
-* Moved `export` statements to the top of the file in `dict_list.jl`
 # 2.10.0
 * Add `DRWATSON_UPDATE_MSG` ENV VAR for suppressing update messages
 # 2.9.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 2.10.1
+# 2.11.0
 * Add `Derived` type and function to allow `dict_list` to compute some parameters from the value of other parameters.
 * Moved `export` statements to the top of the file in `dict_list.jl`
 # 2.10.0
--- a/docs/src/real_world.md
+++ b/docs/src/real_world.md
@@ -289,7 +289,7 @@ general_args2 = Dict(
     "noise2" => [1.0, ComputedParameter(["noise", "N"], (x,y) -> 2x + y)],
     "noisy_training" => true,
     "N" => 100,
-       )
+)
 ```
 ```@example customizing
 dicts2 = dict_list(general_args2)
--- a/docs/src/real_world.md
+++ b/docs/src/real_world.md
@@ -279,6 +279,24 @@ dicts = dict_list(general_args)
 println("Total dictionaries made: ", length(dicts))
 dicts[1]
 ```
+Also, using the type [`ComputedParameter`](@ref), we can have parameters that are computed depending on the value of other parameters:
+```@example customizing
+using DrWatson
+
+general_args2 = Dict(
+    "model" => "barkley",
+    "noise" => [0.075, 0.050, 0.025],
+    "noise2" => [1.0, ComputedParameter(["noise", "N"], (x,y) -> 2x + y)],
+    "noisy_training" => true,
+    "N" => 100,
+       )
+```
+```@example customizing
+dicts2 = dict_list(general_args2)
+println("Total dictionaries made: ", length(dicts2))
+dicts2[1]
+```
+
 Now, how you use these dictionaries is up to you. Typically each dictionary is given to a `main`-like Julia function which extracts the necessary data and calls the necessary functions.
 
 Let's say I have written a function that takes in one of these dictionaries and saves the file somewhere locally:
--- a/docs/src/real_world.md
+++ b/docs/src/real_world.md
@@ -279,7 +279,7 @@ dicts = dict_list(general_args)
 println("Total dictionaries made: ", length(dicts))
 dicts[1]
 ```
-Also, using the type [`ComputedParameter`](@ref), we can have parameters that are computed depending on the value of other parameters:
+Also, using the type [`Derived`](@ref), we can have parameters that are computed depending on the value of other parameters:
 ```@example customizing
 using DrWatson
 
--- a/src/dict_list.jl
+++ b/src/dict_list.jl
@@ -77,11 +77,11 @@ function dict_list(c::AbstractDict)
                                  end
                                  Dict([k=>lookup_candidate(c,trial,k) for k in keys(trial)])
                              end)
-        return produce_computed_parameters(collect(filter(parameter_sets) do trial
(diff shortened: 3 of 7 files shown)


aramirezreyes:
This is a prototype that I thought of today. To date it does not break any test. 
So far, a ComputedParameter can only depend on 1 pre-existing parameter.

Does this design make sense @Datseris?

Datseris:
This does make sense to me! But also, shouldn't we allow the produced parameter to depend in as many parameters as the user wants...?

Just make the field of the parameter a vector of values instead, and if a user gives a single string/symbol make it into a vector anyways.

aramirezreyes:
I added this change and some docs. Now the use cases I added to the tests are:

```julia

p = Dict(:α => [1, 2],
    :solver => [SolverA,SolverB],
    :β => ComputedParameter(:α, x -> x^2),
    )


@test Set([ Dict(:α => 1, :solver => SolverA, :β => 1),
    Dict(:α => 2, :solver => SolverA, :β => 4),
    Dict(:α => 1, :solver => SolverB, :β => 1),
    Dict(:α => 2, :solver => SolverB, :β => 4),
    ]) == Set(dict_list(p))

p2 = Dict(:α => [1, 2],
    :β => [10,100],
    :solver => [SolverA,SolverB],
    :γ => ComputedParameter([:α,:β], (x,y) -> x^2 + 2y),
    )

@test Set([ Dict(:α => 1, :solver => SolverA, :β => 10, :γ => 21),
    Dict(:α => 2, :solver => SolverA, :β => 10, :γ => 24),
    Dict(:α => 1, :solver => SolverB, :β => 10, :γ => 21),
    Dict(:α => 2, :solver => SolverB, :β => 10, :γ => 24),
    Dict(:α => 1, :solver => SolverA, :β => 100, :γ => 201),
    Dict(:α => 2, :solver => SolverA, :β => 100, :γ => 204),
    Dict(:α => 1, :solver => SolverB, :β => 100, :γ => 201),
    Dict(:α => 2, :solver => SolverB, :β => 100, :γ => 204),
   ]) == Set(dict_list(p2))
```

Datseris:
This is great! You should
0. Consider renaming the whole `ComputedParameter` to just `Derived`. Everything is computed after all :) 
1. Increase minor version in Project.toml
2. Add a changelog entry for this
3. Add a refenrece to `Derived` in the docstring of `dict_list`.
4. Actually add `Derived` to the documentaiton, which means go to the file `run&list.md`, and in the `@docs` block that has `dict_list`, also add `Derived`.

then I'll merge and tag!

src/dict_list.jl
```diff
@@ -275,3 +275,82 @@ julia> dict_list(d) # only in case `:a` is `1` the dictionary will get key `:c`
...
+
+"""
+    ComputedParameter(independentP <: Union{String,Symbol}, func :: Function)
+Constructs a ComputedParameter from a single independent parameter.
+"""
+function ComputedParameter(independentP :: Union{String,Symbol}, func :: Function) 
+    return ComputedParameter([independentP], func)
+end
+
+export ComputedParameter
```
Datseris, line 340: Also please move this to the top of the file with the other export statements

src/dict_list.jl
```diff
@@ -275,3 +275,82 @@ julia> dict_list(d) # only in case `:a` is `1` the dictionary will get key `:c`
...
+
+"""
+   produce_computed_parameter(dicts)
+Receive an array of parameter dictionaries, and for each one, evaluates the computed parameters after the possible combination of parameters has been created.
+"""
+function produce_computed_parameters(dicts)
+    map(dicts) do dict
```
Datseris, line 347: By the way, we shouldn't `map` here. This creates a container that we don't need. We should simple loop over the `dict` in `dicts` and perform the subsequent code!

src/dict_list.jl
aramirezreyes: Oh yes. Actually, this file did not have the exports at the top. I took the liberty of moving them up.

src/dict_list.jl
aramirezreyes: Fixed!

aramirezreyes:
I think everything is addressed!

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.11.0
+* Add `Derived` type and function to allow `dict_list` to compute some parameters from the value of other parameters.
+* Moved `export` statements to the top of the file in `dict_list.jl`
```
Datseris, line 3:
```suggestion

```
This has no reason to be mentioned in the changelog xD

docs/src/real_world.md
```diff
@@ -279,6 +279,24 @@ dicts = dict_list(general_args)
 println("Total dictionaries made: ", length(dicts))
 dicts[1]
 ```
+Also, using the type [`ComputedParameter`](@ref), we can have parameters that are computed depending on the value of other parameters:
```
Datseris, line 282:
```suggestion
Also, using the type [`Derived`](@ref), we can have parameters that are computed depending on the value of other parameters:
```

src/dict_list.jl
```diff
@@ -275,3 +298,85 @@ julia> dict_list(d) # only in case `:a` is `1` the dictionary will get key `:c`
...
+struct Derived{T}
+    independentParam::Vector{T}
+    func::Function
+end
+
+"""
+    Derived(independentP <: Union{String,Symbol}, func :: Function)
+Constructs a Derived from a single independent parameter.
+"""
+function Derived(independentP :: Union{String,Symbol}, func :: Function)
```
Datseris, line 362: weird syntax to have white space between `::`. I/ve never seen it before. Should be without white space.

src/dict_list.jl
```diff
@@ -275,3 +298,85 @@ julia> dict_list(d) # only in case `:a` is `1` the dictionary will get key `:c`
      end
      :(toDependentParameter($(esc(value)),$(esc(condition))))
  end
+
+"""
+   Derived(parameters::Vector{Union{String,Symbol}},function::Function)
```
Datseris, line 303:
```suggestion
   Derived(parameters::Vector{Union{String,Symbol}}, function::Function)
```

src/dict_list.jl
```diff
@@ -275,3 +298,85 @@ julia> dict_list(d) # only in case `:a` is `1` the dictionary will get key `:c`
...
+Wrap the name(s) of a parameter(s) and a function. After the 
+possible parameter combinations are created, dict_list will replace instances of
```
Datseris, line 305:
```suggestion
possible parameter combinations are created, [`dict_list`](@ref) will replace instances of 
```

src/dict_list.jl
```diff
@@ -275,3 +298,85 @@ julia> dict_list(d) # only in case `:a` is `1` the dictionary will get key `:c`
...
+Constructs a Derived from a single independent parameter.
+"""
+function Derived(independentP::Union{String,Symbol}, func::Function) 
+    return Derived([independentP], func)
+end
+
+
+"""
+   produce_computed_parameter(dicts)
+Receive an array of parameter dictionaries, and for each one, evaluates the computed parameters after the possible combination of parameters has been created.
```
Datseris, line 369:
```suggestion
Receive an array of parameter dictionaries, and for each one, evaluate
the computed parameters after the possible combination of 
parameters has been created.
```

docs/src/real_world.md
```diff
@@ -279,6 +279,24 @@ dicts = dict_list(general_args)
...
+```@example customizing
+using DrWatson
+
+general_args2 = Dict(
+    "model" => "barkley",
+    "noise" => [0.075, 0.050, 0.025],
+    "noise2" => [1.0, ComputedParameter(["noise", "N"], (x,y) -> 2x + y)],
+    "noisy_training" => true,
+    "N" => 100,
+       )
```
Datseris, line 292:
```suggestion
)
```

Datseris:
thanks, I Think that's a great feature :)

aramirezreyes:
Thanks!

Issue #316, opened 2022-01-11
Workflow documentation for experimentalists


stillyslalom:
I've been using DrWatson.jl to organize my preprocessing code/analysis for an ensemble of experiments performed in a large, complex facility. I've struggled to find an ergonomic workflow that captures the entire pipeline. Issues include:

* Too many experiment-relevant settings to capture in `savename` format without exceeding filesystem length limits
* Experimental ensembles are collected using the same nominal conditions for each data series, but contain idiosyncrasies (inoperable instruments, flubs, sensor/timing tweaks) that may need to be handled on a case-by-case basis
* External tools (ImageJ, Matlab) and user input fit awkwardly in the middle of a `produce_or_load` data-processing pipeline that assumes hands-off, end-to-end Julia code
* Data in instrument-specific formats is best kept in the filesystem where it can be handled with specialized external software rather than packed into JLD

This issue is partly a reminder to myself to write suitable documentation once I arrive at a good workflow.

Datseris:
@stillyslalom can you try this out and let us know how it works for you? @sebastianpech I suggest that we start with writing a "real world example" in the DrWatson docs that showcases (briefly of course) your workflow. Based on that perhaps we can think of integrating DrWatsonSim directly into DrWatson if possible? Seems like many other people have asked for something similar and it always goes back to something like your approach...

sebastianpech:
@Datseris Yes, seems so. I'll open a PR and polish the code a bit. I think I kept it general enough to integrate it in DrWatson. And it's all opt-in features, so no existing workflows will break and the metadata directory is only created when a metadata-related function is called the first time, so the folder structure also remains unchanged.

sebastianpech:
Oh and @Datseris which file format do we prefer. The version on the master branch uses BSON for storing metadata, but I also have a separate branch using JLD2. Haven't merged it yet as I didn't want to update all my old project folders, but I guess as we have JLD2 as a dependency in DrWatson, I will do the switch.

Datseris:
Yeah there is a clear statement in DrWatson now for JLD2 preference since 2.0. Do note that JLD2 cannot save functions though. Do you need this?

sebastianpech:
Aha! I remember why I didn't do the switch yet. Yes I store all kinds of stuff in the bson files. However, it's not a requirement for the metadata functions to work. I think it's easy to workaround this. I will add a note to the docs.

Datseris:
[...]

PR #366, opened 2022-11-02
Allow arbitrary functions to extract name from configuration in `produce_or_load`

Datseris:
Alright, so this PR kinda solves my problems I had with really wanting to use `produce_or_load`, but also having too many, or too complicated, simulation parameters. I guess it is a simple inbetween step until DrWatson has some proper way to manage metadata/configuration "files". The discussion here https://github.com/JuliaDynamics/DrWatson.jl/issues/316 and the PR https://github.com/JuliaDynamics/DrWatson.jl/pull/333 are relevant.

My intermediate solution is to allow arbitrary functions in `produce_or_load` that take as an input the `config` container, and return a hopefully unique string for it. In my personal workflow I am using `objectid`, but I haven't tested it extensively to see really how unique or safe its output is. When DrWatsonSim functionality is in, the change in this PR would still allow us to effortlessly integrate DrWatsonSim with `produce_or_load`.

See the new example added in the Real World Usage for an overview of why this functionality is useful: https://juliadynamics.github.io/DrWatson.jl/previews/PR366/real_world/#produce_or_load-with-hash-codes-1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2.12.0
+- Arbitrary functions extracting strings from data can be used in `produce_or_load` instead of `savename`. `hash` is the most useful function here. An example in Real World Examples highlights this.
+- Additional keywords in `produce_or_load` propagated to `savename` are deprecated.
+- The example project created in the Workflow Tutorial no longer has spaces in its name
+
 # 2.11.0
 - Now the default project with `initialize_project` will include a documentation and a test folder, and setup CI for both automatically.
 - `@produce_or_load` was broken but because CI was disabled this was silently ignored. `produce_or_load` now has call signature `produce_or_load(f::Function, config, path::String = "")` and same signature for the macro with `path` mandatory.
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.11.1"
+version = "2.12.0"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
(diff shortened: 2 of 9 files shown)


Datseris:
Ah no my dreams are ruined. Seems like this isn't reliable. Here:
```julia
julia> objectid(rand(Random.MersenneTwister(4321), 1000))
0xc8f87467e4da680a

julia> objectid(rand(Random.MersenneTwister(4321), 1000))
0x0205f59056f4c6e6

julia> objectid(rand(Random.MersenneTwister(4321), 1000))
0xd49ed4947c3f3987
```

@sebastianpech this won't work unfortunately. Is my dream of using `produce_or_load` with practically everything going down the drain...? Bu hu hu.

Datseris:
MY DREAM IS SAVED I JUST NEEDED HASH:
```julia
julia> hash(rand(Random.MersenneTwister(4321), 1000))
0xc3437f37a6e8d226

julia> hash(rand(Random.MersenneTwister(4321), 1000))
0xc3437f37a6e8d226

julia> hash(rand(Random.MersenneTwister(4321), 1000))
0xc3437f37a6e8d226
```

Datseris:
This is good to go.

Datseris:
Hm, no, this isn't good to go :( From the example here: https://juliadynamics.github.io/DrWatson.jl/previews/PR366/real_world/#produce_or_load-with-hash-codes-1  we see that once I run the code with exactly the same configuration as it has been run before, e.g., `config = Dict("x" => rand(Random.MersenneTwister(1234)), "f" => f1)`, a new file is produced nevertheless. Why...? `hash` doesn't seem to return the same number even though I give it the same dictionary with exactly the same entries.

Datseris:
Alright, update: I've fixed everything here. First of all, you can use `hash` on functions. But it only utilizes the function name information, which is kinda useless. You cannot use it at all with anonymous functions. I've updated the example to not use functions in general, as it feels like a risky thing to do, and I also added a warning message.

JonasIsensee:
This looks good to me!

PR #376, opened 2022-12-17
Make default gitignore only ignore folders at the top level

KnutAM:
The current default gitignore ignores all folders and files with names in "Folders to ignore"
This PR suggests only ignoring those names if occurring at the top level.


Datseris:
Oh yes absolutely, this was the intention from the start. Could you please also increment patch version in the Project.toml file? I count this as a bug fix!

PR #380, opened 2023-03-13
Fix to Issue#181

### LINKED ISSUE #181: Handle trailing slashes in collect_results!
When the `folder` argument to `collect_results!` contains a trailing slash (e.g `/home/lhupe/coolresults/`), the default setting for `filename` breaks. The result file will be saved as `/home/lhupe/coolresults/results_.bson`. This has the unfortunate side effect that on the next run, `collect_results!` will create an entry for its own cache file.
While this is not a bug of `DrWatson `(the issue is with `Base.splitdir`, which naively splits paths on the final separator), it would be nice to account for this in the default definition of `filename` (e.g. using `rstrip(folder, '/')` to remove trailing slashes).

Myrkwid:
This is my fix #181.
Using the mentioned `rstrip()` for the creation of the default value of `filename` in `result_colletion.jl` the described error will no longer occur.
By also adding `rstrip()` to `projectdir()` function, the error is also fixed if `collect_results!()` is called using defined variables using `datadir()` (as is done in `update_results_tests.jl`).
To demonstrate that, I included a test within `update_resluts_tests.jl` that checks if the bugged file is created.
In the case that the wrongly named and placed file is created, the test will remove it anyways, since it would interfere with the test regarding the dimensions of the dataframes (e.g. ` @test size(cres2) == (6, 7)`).

I hope you will find this helpful!

### DIFF
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -29,9 +29,9 @@ function projectdir()
     if is_standard_julia_project()
         @warn "Using the standard Julia project."
     end
-    dirname(rstrip(Base.active_project(), '/'))
+    dirname((Base.active_project()))
 end
-projectdir(args...) = joinpath(projectdir(), args...)
+projectdir(args...) = rstrip(joinpath(projectdir(), args...), '/')
 
 
 # Generate functions to access the path of default subdirectories.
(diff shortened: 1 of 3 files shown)


Datseris (approved):
We do find this helpful! Thanks!

Issue #368, opened 2022-11-09
GitReference should be LibGit2.GitReference


NicholasWMRitchie:
I don't think that LibGit2 exports GitReference.
GitReference is used on line 308 of project_setup.jl.

`LibGit2.delete_branch(GitReference(repo, "refs/heads/$default"))` should probably be `LibGit2.delete_branch(LibGit2.GitReference(repo, "refs/heads/$default"))`

Datseris:
ah thats why this doesnt work... can you open a PR?

PR #381, opened 2023-03-13
Fix for Issue#251

### LINKED ISSUE #251: Have initialize_project() initialize the git repo with a `main` branch instead of `master`.
Have `initialize_project()` initialize the git repo with a **main** branch instead of **master**.

The git command is simply `git branch -m master main`.

Still learning Julia, so I'm not quite sure how to make a PR with an edit to the source.

While it's true a user can configure their git to default to **main** with `git config --global init.defaultBranch main`, this would be a beneficial change for DrWatson to have by default until the baked in default for git is **main**.

Myrkwid:
This is my fix #251.

As far as I can tell the problem was that [GitReference()](https://github.com/JuliaDynamics/DrWatson.jl/blob/main/src/project_setup.jl#L308) needed the prefix `LibGit2.`.

This fixes the problem (for me) and the warning message no longer occurs.
The created project only has 1 branch named `main` even if global settings of the user are different.

### DIFF
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -305,7 +305,7 @@ function initialize_project(path, name = default_name_from_path(path);
             branch = "main"
             if branch != default
                 LibGit2.branch!(repo, branch)
-                LibGit2.delete_branch(GitReference(repo, "refs/heads/$default"))
+                LibGit2.delete_branch(LibGit2.GitReference(repo, "refs/heads/$default"))
             end
         catch err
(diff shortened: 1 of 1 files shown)


Datseris:
thank you :) this has been bugging me for a while now...

BeastyBlacksmith:
That should alsofix https://github.com/JuliaDynamics/DrWatson.jl/issues/368 then, no?

PR #383, opened 2023-03-15
Force convert of projectdir to String

### LINKED ISSUE #382: v2.12.2 breaks compatibility of produce_or_load() and datadir()
**Describe the bug**
When using `produce_or_load(datadir("sims"), myparams, myfun)`  in v2.12.2 (.1 works fine) I get the error
```
MethodError: no method matching produce_or_load(::SubString{String}, ::Dict{String, Any}, ::typeof(myfun))
```

It seems that a String instead of the SubString would work.
Maybe related to #380 ?

**Minimal Working Example**
Sorry, I have no MWE yet.

```
julia> versioninfo()
Julia Version 1.8.5
Commit 17cfb8e65e (2023-01-08 06:45 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, tigerlake)
  Threads: 8 on 8 virtual cores
Environment:
  JULIA_EDITOR = code
```

Tom271:
Converts output of `projectdir` to String. Added test for all *dir() functions whish is redundant when testing `projectdir()` directly. Happy to remove the extra test if you think it's unnecessary. Fixes #382. 

Passes all tests locally

### DIFF
--- a/src/project_setup.jl
+++ b/src/project_setup.jl
@@ -31,7 +31,7 @@ function projectdir()
     end
     dirname(Base.active_project())
 end
-projectdir(args...) = rstrip(joinpath(projectdir(), args...), '/')
+projectdir(args...) = String(rstrip(joinpath(projectdir(), args...), '/'))
 
 
 # Generate functions to access the path of default subdirectories.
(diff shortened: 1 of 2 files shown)


Datseris:
thanks a  lot, sorrry for the hickup....

PR #389, opened 2023-05-08
Update result_collection.jl

yp51md:
add `verbose &&` before all the `@info` for a clean print.


Datseris:
great, also bump patch version (3rd number) in the version in Project.toml file

Issue #392, opened 2023-06-30
produce_or_load doubles filename prefixes


pjgorski:
**Describe the bug**
When a `config` variable `default_prefix` is modified, then `produce_or_load` saves the output to a file with a name starting with two prefixes instead of one. 

**Minimal Working Example**
```
using DrWatson
using Dates

quickactivate(@__DIR__)

mutable struct Params
    A::Int 
    B::Float64 
    C::String 
    D::Array
    date::Date
end

function Params(;
    A = 36,
    B = 10.,
    C = "test",
    D = rand([1,2,3,5], 5)
)
    t_ = Dates.now()
    return Params(A, B, C, D, Date(t_))
end

mutable struct Params2
    A2::Int 
    B2::Float64 
    C2::String 
    D2::Array
    date::Date
end

function Params2(;
    A2 = 36,
    B2 = 10.,
    C2 = "test",
    D2 = rand([1,2,3,5], 5)
)
    t_ = Dates.now()
    return Params2(A2, B2, C2, D2, Date(t_))
end


DrWatson.default_prefix(params::Params) = "Experiment_"*string(params.date) 
DrWatson.allaccess(::Params) = (:A, :B, :C, :D)
DrWatson.allaccess(::Params2) = (:A2, :B2, :C2, :D2)


params = Params(A = 36,B = 10., C = "test", D = rand([1,2,3,5], 5))
params2 = Params2(A2 = 36, B2 = 10., C2 = "test", D2 = rand([1,2,3,5], 5))

what_to_save = produce_or_load(projectdir("test", "produce_or_load_test_files"), params; 
    force = false, loadfile = true) do params
        "some test output"
        Dict("A" => rand(5), "B" => rand(1:5, 5))
end

what_to_save = produce_or_load(projectdir("test", "produce_or_load_test_files"), params2; 
    force = false, loadfile = true) do params2
        "some test output"
        Dict("A" => rand(5), "B" => rand(1:5, 5))
end
```
As a result following files are saved: `A2=36_B2=10.0_C2=test.jld2` and `Experiment_2023-06-30_Experiment_2023-06-30_A=36_B=10.0_C=test.jld2`.
The first file is the result of saving `Params2` for which `default_prefix` was not modified. The second file has the prefix repeated. 

**Versions**
DrWatson 2.12.5
Julia 1.6.4

**Proposed solution**
I propose moving line 91 in file `saving_files.jl` to line 87, and changing line 83, so they would be:
```
[...]

PR #393, opened 2023-06-30
Removing double prefixes bug in produce_or_load

pjgorski:
See #392 

I also created a test, but I was unable to test it locally. Each time I ran tests `BSON` package was uninstalled, and the lack of this package produced errors, and further tests were unable to be run. The issue might be with the system not making a distinction between a system package and a local one.

### DIFF
--- a/test/savefiles_tests.jl
+++ b/test/savefiles_tests.jl
@@ -307,6 +307,18 @@ end
     rm(path; recursive = true, force = true)
 end
 
+# Testing proper filenames when default_prefix was modified. See https://github.com/JuliaDynamics/DrWatson.jl/issues/392
+@testset "@produce_or_load with default_prefix modified" begin
+    path = mktempdir()
+
+    DrWatson.default_prefix(ntuple::NamedTuple) = "Prefix_"
+
+    sim, path = produce_or_load(f, simulation, "")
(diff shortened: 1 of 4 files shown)


test/savefiles_tests.jl
```diff
@@ -307,6 +307,18 @@ end
...
 end
 
+# Testing proper filenames when default_prefix was modified. See https://github.com/JuliaDynamics/DrWatson.jl/issues/392
+@testset "@produce_or_load with default_prefix modified" begin
+    path = mktempdir()
+
+    DrWatson.default_prefix(ntuple::NamedTuple) = "Prefix_"
+
+    sim, path = produce_or_load(f, simulation, "")
+    @test path == savename(simulation)
```
Datseris, line 317: Please don't infect the `NamedTuple` here, because this could mess up tests after this one (if some are added in the future). Instead make a dummy type like
```julia
struct Dummy
x
y
end

simulation = Dummy(1,2)
DrWatson.default_prefix(...)
```

Datseris (changes requested):
Thanks for the quick fix. I think the test can be set up better.

pjgorski:
I managed to run and pass the tests locally.

Issue #387, opened 2023-04-27
tag with git commit message


harrybooth:
Maybe I have overlooked this functionality, but would it be possible to add the git commit message to the result dictionary using [DrWatson.@tag!](https://juliadynamics.github.io/DrWatson.jl/v0.8/save/#DrWatson.@tag!)?

This would be really useful, as for example the git commit hash e.g. gitcommit => "96df587e45b29e7a46348a3d780db1f85f41de04" is non-descriptive, especially if you are trying to compare different versions of the same data and understand what the change was. Allowing the git commit message to be included, e.g. "changed parameter x to y" would be a nice automated way to relate code changes to experimental results.

Datseris:
Sure, this seems like providing a keyword arugment `use_description = false` to `tag!`. If `true`, the commit description is used instead of the tag. Note that this is dangerous: commit descriptions can be arbitrarily small and non-unique, so this may not really help you reverse time and go back to given commit. But of course use at your own risk hence the default value will be `false`.

Datseris:
sure

Datseris:
we can do it like the `gitpatch` keyword: the commit message can be added as an additional field instead of replacing or mingling with the git commit field

PR #398, opened 2023-09-12
#387

Cia15:
This is my proposed fix for the issue #387. 
Using LibGit2.GitCommit() and LibGit2.message() I added the gitcommit message. However, I haven't written the test for this one. Any feedback would be welcomed. Have a nice day!


Datseris (commented):
Why do we need the statistics dependency and the `settings.json` file?

For the tests: you said you haven't written any tests, but did you try out this PR locally and it did work, i.e, it solved the #387 ?

src/saving_tools.jl
```diff
@@ -112,10 +112,7 @@ function read_stdout_stderr(cmd::Cmd)
     return (exception = exception, out=read(out,String), err=read(err,String))
 end
 
-"""
-    gitpatch(gitpath = projectdir())
-
-Generates a patch describing the changes of a dirty repository
+""" Generates a patch describing the changes of a dirty repository
```
Datseris, line 115: I am confused about this documentation string change. It is a mistake. The previous version was obviously clearer.

src/saving_tools.jl
```diff
@@ -200,6 +199,7 @@ Dict{Symbol,Any} with 3 entries:
 function tag!(d::AbstractDict{K,T};
         gitpath = projectdir(), force = false, source = nothing,
         storepatch::Bool = readenv("DRWATSON_STOREPATCH", false),
+        mssg::Bool = false
```
Datseris, line 202: `mssg` is not a clear keyword. PLease name it something that conveys its purpose better just as `commit_message`

Datseris:
In any case, you _need_ to write tests for this PR, by going into the `test` folder, and modifying one of the existing tests. Pass the keyword in, the load the file, and then see that it has a commit message as field.

Datseris:
And also, thank you for the contribution!

Cia15:
Dear George Datseris,

For preface I would like to apologize for some of the unclear/weird changes
on my part. I was doing the issue for a project for my university module
and I have limited knowledge and experience with working with github or
fixing issue in general.

To address some of your feedback:
1. We don't need statistics dependency and the settings.json file, it was a
mistake on my part. When we run it locally, it did work and shows the git
commit message.
2. The documentation string change is also my fault on that part.
3. The mssg was chosen because my instructor and I thought it would clash
with the message function? However, thank you so much for the input.
4. I will try to add the test.

Once again thank you so much for the feedback!

Cheers,
Cia15

George Datseris ***@***.***> schrieb am Mi., 13. Sept. 2023,
6:27 PM:

src/saving_tools.jl
Cia15: Alright, thank you for your feedback. I already changed the mssg keyword to commit_message

test/stools_tests.jl
```diff
@@ -35,7 +35,7 @@ cpath = _setup_repo(false) # clean
 @test !endswith(gitdescribe(cpath), "-dirty")
 
 # tag!
-function _test_tag!(d, path, haspatch, DRWATSON_STOREPATCH)
+function _test_tag!(d, path, haspatch, has_message, DRWATSON_STOREPATCH)
```
Datseris, line 38: Its great that you modified this function to test for the existence of a message. However I see two problems: 1) you never call this function with `true` as the 4th argument, and hence, the function never actually tests the new code. 2) the function does not test whether the message itself is correct. Only tests that it is not the empty string.

I propose for you to _revert_ the modification of the `test_tag!` function. Instead, in the codeblock `@testset "message_name" begin` that you have created below, write a custom, hand coded test that explicitly calls `tag!` and then it explicitly tests that the tagged dictionary has exact information of the commit message, whatever that message is.

.vscode/settings.json
Datseris, line 1: The `settings.json` file still exists.

src/saving_tools.jl
```diff
@@ -112,9 +112,7 @@ function read_stdout_stderr(cmd::Cmd)
     return (exception = exception, out=read(out,String), err=read(err,String))
 end
 
-"""
-    gitpatch(gitpath = projectdir())
-
+"""
```
Datseris, line 115: Why did this change happen? You should revert it.

src/saving_tools.jl
```diff
@@ -128,6 +126,7 @@ function gitpatch(path = projectdir(); try_submodule_diff=true)
         repo = LibGit2.GitRepoExt(path)
         gitpath = LibGit2.path(repo)
         gitdir = joinpath(gitpath,".git")
+        ###gitmessage = LibGit2.message() #Add the gitmessage
```
Datseris, line 129: This comment does not help, rather, it confuses. You should delete it.

src/saving_tools.jl
```diff
@@ -200,6 +200,7 @@ Dict{Symbol,Any} with 3 entries:
 function tag!(d::AbstractDict{K,T};
         gitpath = projectdir(), force = false, source = nothing,
         storepatch::Bool = readenv("DRWATSON_STOREPATCH", false),
+        commit_message::Bool = false
```
Datseris, line 203: Note that you have added this new functionality but you have not _documented_ it yet. You need to edit the `tag!` function documentation string and describe this new functionality.

src/saving_tools.jl
```diff
@@ -208,6 +209,7 @@ function tag!(d::AbstractDict{K,T};
     # Get the appropriate keys
     commitname = keyname(d, :gitcommit)
     patchname = keyname(d, :gitpatch)
+    message_name = keyname(d, :gitmessage) ### added
```
Datseris, line 212: What information does the `### added` comment provide? Does it really help? Remove the comment.

src/saving_tools.jl
```diff
@@ -222,6 +224,14 @@ function tag!(d::AbstractDict{K,T};
                 d[patchname] = patch
             end
         end
+        if commit_message
+            repo = LibGit2.GitRepoExt(gitpath)
+            mssgcommit =  LibGit2.GitCommit(repo, "HEAD")
+            msg = LibGit2.message(mssgcommit)
+            if (msg !== nothing) && (msg != "")
+                 d[message_name] = msg
+            end ### added
```
Datseris, line 233: again, remove `### added`.

test/stools_tests.jl
Cia15: I see, thank you! I will try to implement it

src/saving_tools.jl
```diff
@@ -112,8 +112,8 @@ function read_stdout_stderr(cmd::Cmd)
     return (exception = exception, out=read(out,String), err=read(err,String))
 end
 
-"""
-    gitpatch(gitpath = projectdir())
+""" 
+gitpatch(gitpath = projectdir())
```
Datseris, line 116: Incorrect. Restore the 4 spaces of identation, otherwise the documentation string will not register correctly.

src/saving_tools.jl
```diff
@@ -167,7 +168,8 @@ the project's gitpath). Do nothing if a key `gitcommit` already exists
 repository is not found. If the git repository is dirty, i.e. there
 are un-commited changes, and `storepatch` is true, then the output of `git diff HEAD` is stored
 in the field `gitpatch`.  Note that patches for binary files are not
-stored. You can use [`isdirty`](@ref) to check if a repo is dirty.
+stored. You can use [`isdirty`](@ref) to check if a repo is dirty. If the commit message is set to true, 
+then the output will display the commit message.
```
Datseris, line 172: When referring to code, don't use english words. Use code, formatted as monotype font. Don't write "If the commit message is ...". Write instead "If the `commit_message` keyword is set to `true`, then...".

src/saving_tools.jl
```diff
@@ -190,16 +192,18 @@ Dict{Symbol,Int64} with 2 entries:
   :y => 4
   :x => 3
 
-julia> tag!(d)
+julia> tag!(d, commit_message=true)
```
Datseris, line 195:
```suggestion
julia> tag!(d; commit_message=true)
```
it is better syntax to separate keywords by `;`.

test/stools_tests.jl
```diff
@@ -56,18 +56,32 @@ d2 = Dict("x" => 3, "y" => 4)
...
+        message_commit_test = LibGit2.GitCommit(repo, "HEAD")
+        message_name = LibGit2.message(message_commit_test)
+        if (message_name !== nothing) && (message_name != "")
+            print(message_name)
+       end
     end
 end
 
+# Ensure that the correct git message show up
+d_new = tag!(d, gitpath=dpath, commit_message = true)
```
Datseris, line 82: This doesn't test anything, see above.

test/stools_tests.jl
```diff
@@ -56,18 +56,32 @@ d2 = Dict("x" => 3, "y" => 4)
...
+    end
+    @testset "message_name" begin
+        path = mktempdir(cleanup=true) # delete path on process exit
+        repo = LibGit2.init(path)
+        LibGit2.commit(repo, "tmp repo commit")
+        message_commit_test = LibGit2.GitCommit(repo, "HEAD")
+        message_name = LibGit2.message(message_commit_test)
+        if (message_name !== nothing) && (message_name != "")
+            print(message_name)
+       end
```
Datseris, line 77: This test block is wrong unfortunately. **You need to test the `tag!` function. This is what DrWatson provides. We don't care about testing Lib2Git.**

Please call the `tag!` function, with the keyword `commit_message` to true. Then manually examine the tagged dictionary, and manually write the test 
```
@test d["gitmessage"] == "the mesage you expecet"
```
Not calling the `@test` macro means you are not testing anything. Printing is not testing.

Datseris (changes requested):
I am sorry it took so long to submit another review. I was overwhelemd with other projects.

I am also sorry to drag this PR more, but the testing is still incorrect and needs to change. I am positive that this will be the last round of review though, I tried to make things explicit!

test/stools_tests.jl
Cia15: Sorry, this might be a dumb question, but will it also work if instead I add the `@test d["gitmessage"] == "the mesage you expect"` in the `_test_tag!` function?  So something like this? I'm terribly sorry for the beginner mistakes..

image

Datseris: Don't alter any of the previous code. It is simply not worth it. Why not just add this new test code block, that just has 5 lines of code, and does the test? Simpler = better.

(Sure, there could be ways to alter the previous code, but why do it though? just make a new `@testset` codeblock with 4 lines of code that call the `tag!` function and read the message and call an `@test` on the read message.)

Cia15: Somehow it didn't work for me.............

src/saving_tools.jl
```diff
@@ -167,7 +168,8 @@ the project's gitpath). Do nothing if a key `gitcommit` already exists
 repository is not found. If the git repository is dirty, i.e. there
 are un-commited changes, and `storepatch` is true, then the output of `git diff HEAD` is stored
 in the field `gitpatch`.  Note that patches for binary files are not
-stored. You can use [`isdirty`](@ref) to check if a repo is dirty.
+stored. You can use [`isdirty`](@ref) to check if a repo is dirty. If the `commit message` is set to `true`,
```
Datseris, line 171:
```suggestion
stored. You can use [`isdirty`](@ref) to check if a repo is dirty. 
If the keyword `commit_message` is set to `true`, 
```

it is important to be _precise_ in documentation strings.

Datseris (changes requested):
Besides the outlined changes, you need to also:

- Got to Project.toml and increment the minor (2nd number) version.
- Got to CHANGELOG.md and add a new section under the new minor version where you state that you added this new functionality.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DrWatson"
 uuid = "634d3b9d-ee7a-5ddf-bec9-22491ea816e1"
 repo = "https://github.com/JuliaDynamics/DrWatson.jl.git"
-version = "2.12.7"
+version = "2.13.7"
```
```suggestion
version = "2.13.0"
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.13.7
```
Datseris, line 1:
```suggestion
# 2.13.0
```

Datseris (approved):

Issue #403, opened 2024-02-05
Customize folders to be git-ignored


fabiofortkamp:
**Is your feature request related to a problem? Please describe.**
I do not always agree with DrWatson's default options for the .gitignore file. I extensively use Quarto notebooks, which are leightweight enough to be included in version control, and my data files usually are small CSV and text files that can be included in the repository, so that everything is ready when the project is clone. So I always have to exclude the `notebooks` and `data` folders in the `.gitignore` file.

**Describe the solution you'd like**
I think that this file could be customized during project creation. 

*I actually have a working solution that modifies the `initialize_project` function with a new keyword option that accepts a vector of folders to be ignored (the default is everything that is currently in .gitignore). Should I submit a PR?*

**Describe alternatives you've considered**
I guess the alternative would be to comply with the expected behavior and create a script to download the data files and host the notebooks in some external services, calling the functions in the DrWatson project I'm developing. For my small projects, this seems more trouble that it's worth.

Datseris:
> I actually have a working solution that modifies the `initialize_project` function with a new keyword option that accepts a vector of folders to be ignored (the default is everything that is currently in .gitignore). Should I submit a PR?

yes please, this seems to me a valid and straightforward improvement!

Datseris:
> **Describe alternatives you've considered**
> I guess the alternative would be to comply with the expected behavior and create a script to download the data files and host the notebooks in some external services, calling the functions in the DrWatson project I'm developing. For my small projects, this seems more trouble that it's worth.

Modifying the `.gitignore` file seems like a much simpler alternative to this, but in any case allowing this at project construction is even better.

PR #404, opened 2024-02-06
Extract function to create custom gitignore file

fabiofortkamp:
This attemps to solve #403 by including a keyword argument to the `initialize_project` function to customize the folders to be included in the `.gitignore` file. Default options is to preserve the current gitignore file.


Datseris (approved):

Datseris:
Thanks, this is an example of a perfect pr: has tests, docs, and increments project version. will merge once tests pass!

PR #414, opened 2024-04-27
Fix downloads count badge

pricklypointer:
Fixed the download count badge using https://juliapkgstats.com/. The current badge is broken because the old API and PkgVisBoard is not maintained.

This PR:
- Replaces the broken badge with a new badge that uses the new API

Please review and provide any feedback. Thanks for considering this PR!


Datseris:
Hi @pricklypointer this works, thank you, but the original badge displayed the _total overall downloads_, not the downloads _per month_. Is it possible to have that statistic instead?

Datseris:
I don't even remember how long DrWtson has been around so I can't even do the averaging myself :D

pricklypointer:
Yes, I will add this API to the site to make it available. I will create a PR when it is up

Datseris:
Thanks a lot, I appreciate it!

PR #415, opened 2024-05-21
Replace monthly download count badge with total downloads badge

pricklypointer:
Ref https://github.com/JuliaDynamics/DrWatson.jl/pull/414


Datseris:
thanks a lot! Do you have any sorts of script to do a mass PR that does this change in many repos?

pricklypointer:
I do not have this, I just manually did it for a couple repos. I will look into creating one

PR #416, opened 2024-06-05
Expand section "making your project a usable module"

### LINKED ISSUE #397: Make a project to be a module by default
I think the root problem is the fact that Revise.jl does not track files that are included in a file that is itself called using includet.
This causes some annoyances when I have a file script.jl file in my script folder and I want to call source code from files in src/.
When I work on a script, I like to include it with includet which allows me to modify the file without reloading it. But then my included files in src/ are not visible. After every modification of the files in src/, you have to restart the REPL which is very annoying. This problem was also described by a user on discourse in [this post ](https://discourse.julialang.org/t/best-debug-workflow-for-dr-watson/97234?u=olivier_merchiers).

The solution I use is the make the project a module as is described in the manual [here](https://juliadynamics.github.io/DrWatson.jl/stable/real_world/#Making-your-project-a-usable-module-1).
This solves all the mentioned issues. It is now possible to add 'using MyProject' in my script files and calling the functions. Revise.jl works perfectly since MyProject is a legal julia project and can hence be loaded. 
The feature I would love to have is that when a project is created, the module would be automatically created. Granted, this is a very small step to perform by hand, but currently the module solution lacks discoverability since:
1.  it is not obvious that this solves a big pain point with the use of includet
2. it is just one of the many possible workflows.

My feature request is hence to make the module by default and advertise this workflow more prominently.
It is not clear for me whether this is breaking or not. I guess for people who don't use the module approach, everything should just work the same. But I do understand that this could be slightly disruptive. 
  

As an alternative, maybe a modification of the documentation would be sufficient to reflect better the advantages of the module approach.

Thanks,
Olivier

abraemer:
Triggered by the [post on discourse](https://discourse.julialang.org/t/drwatson-common-code-as-module-issue/115203) where a user had issues with the conversion of their project into a package, I added a comment to the docs mentioning the need for a `name` in `Project.toml`

I also saw #397 and added some words about the advantages (can use Revise.jl) and disadvantages (frequent recompilation). I did not quite understand why this recompilation should take a long time, since the dependencies should not be recompiled anew AFAIK. But I mentioned it nonetheless.

I did no changes beyond the documentation, so please consider whether this PR fixes #397.

Please read the changes and let me know where I should rephrase things. I tried to stick to the style of the rest of the page.

### DIFF
--- a/docs/src/real_world.md
+++ b/docs/src/real_world.md
@@ -33,8 +33,8 @@ save(datadir("sim", "barkley", "astonishing_results.jld2"), data)
 ```
 
 ## Making your project a usable module
-By default DrWatson.jl's projects are **not** Julia packages since that leads to recompilation of the project every time a source file is changed which can take a substantial amount of time. However, in some cases it can be beneficial nonetheless to turn the project into a usable module, e.g. if you want to use [`Revise.jl`](https://github.com/timholy/Revise.jl) for development. There are also some projects, where it is the case that some packages and files from the source folder are loaded at the beginning of _every file of the project_.
-Let's consider the example, where I have a project that I know that for _any_ script I will write, the first five lines will be:
+For some projects, it is often the case that some packages and files from the source folder are loaded at the beginning of _every file of the project_. 
+For example, I have a project that I know that for _any_ script I will write, the first five lines will be:
 ```julia
 using DrWatson
 @quickactivate "AlbedoProperties"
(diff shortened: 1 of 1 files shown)


docs/src/real_world.md
```diff
@@ -33,8 +33,8 @@ save(datadir("sim", "barkley", "astonishing_results.jld2"), data)
 ```
 
 ## Making your project a usable module
-For some projects, it is often the case that some packages and files from the source folder are loaded at the beginning of _every file of the project_.
-For example, I have a project that I know that for _any_ script I will write, the first five lines will be:
+By default DrWatson.jl's projects are **not** Julia packages since that leads to recompilation of the project every time a source file is changed which can take a substantial amount of time. However, in some cases it can be beneficial nonetheless to turn the project into a usable module, e.g. if you want to use [`Revise.jl`](https://github.com/timholy/Revise.jl) for development. There are also some projects, where it is the case that some packages and files from the source folder are loaded at the beginning of _every file of the project_.
+Let's consider the example, where I have a project that I know that for _any_ script I will write, the first five lines will be:
```
Datseris, line 37: I would revert this back to what it was. Let's keep the discussion here simple. Besides, we don't really discuss packages in this section. The recompilation problem of issue #397 isn't as clear as we thought, so, as you correctly pointed out, this recompilation doesn't happen always, only some particular, difficult to reproduce cases. Hence, we shouldn't say too much abbout this here.

Datseris (approved):

Issue #420, opened 2024-08-11
Add Abiliity to Pass Custom Load Function to `collect_results!`


NuclearPowerNerd:
Firstly, thank you writing this amazing package. I love it and have used it for over a year now in my academic work. 

I have been trying to take full advantage of every capability DrWatson has to offer and I've pretty much done that except for `collect_results!`. I'd really like to be able to take advantage of this because I have many many simulations and a summary table of the most important parameters would really help when preparing results and reviewing what additional runs I need to make.

Here is my problem and I apologize if there is already an obvious solution and I missed it. I save all my simulation results in a struct around which I've built a lot of methods. My structs contain all the parameters that are important to my simulations as well as my results themselves. Something like this

```julia
struct MySim
    Param1::Float64
    Param2::Int64
    Result1::Matrix{Float64}
    # and so on, many more 
end
```

It is important that when I save this to disc, it is saved as a struct and not as a dict. That means that when I save it I do something like this

```julia
wsave("full_path_to_result.jld2", "solution", myresult) # where myresult is of type MySim
```

And when I load I have to do 

```julia
result = wload("full_path_to_result.jld2")["solution"]
```

But this seems problematic for `collect_results!` which expects (as far as I can tell) that `myresult` has passed through `struct2dict`. I really only need this step if I am printing a summary table to file. Thus, I'm wondering can we add a kwarg to `collect_results!` that wraps the call to `wload` just within the source of `collect_results!`? I'm happy to write the pull request but I wanted to make sure this was sensible before I did. I'm thinking something like this.

```julia
function collect_results!(filename, folder;
    valid_filetypes = [".bson", "jld", ".jld2"],
    subfolders = false,
    rpath = nothing,
    verbose = true,
    update = false,
    newfile = false, # keyword only for defining collect_results without !
    rinclude = [r""],
    rexclude = [r"^\b$"],
    load_function = () -> ()
    kwargs...)

# Skipping some code here
data = load_function(wload(filename))
```

This way, I could implement a load function like this and I'd get the desired behavior. 

```julia
load_function = (d) -> struct2dict(d["solution"])
```

Perhaps this is too specialized for my use case? I'm not sure. Please let me know if this could be achieved in a better fashion or if I am missing something here. The only thing I insist on is that I'd like to maintain saving my results as structs and not converting them to dictionaries. If I converted them to dicts I'd have to define new functions to rebuild the structs from the loaded dicts which I'd like to avoid.

[...]

PR #421, opened 2024-08-12
implement custom load function for collect_results

NuclearPowerNerd:
Here is my first pass at implementing #420

This adds an optional wrapper around the loading of data when being processed by `collect_results`. This is useful in case the data being loaded from file needs to be modified before being processed by `collect_results`. The wrapper can be passed to `collect_results` using the new kwarg `load_function_wrapper`.

For example, consider you have a simulation result that is saved to disc as a struct. When loaded from disc it will be as a one-element dict containing that struct. The struct may contain many fields that should form the columns of the DataFrame. In that case you'd want to convert the struct to a dictionary after being loaded but before being processed by `collect_results`. The `load_function_wrapper` provides such a possibility.

By default `load_function_wrapper=nothing` and the default methods will be used to load data (so `wload` and `jld2.jldopen`).

Here is an example usage.

```julia
struct Dummy
    a::Float64
    b::Int64
    c::Matrix{Float64}
end

dummy = Dummy(1.0, 1, rand(3,3))
wsave("dummy.jld2", "dummy", dummy)

tmp = wload("dummy.jld2") # one-element Dict with key "dummy" that contains the struct Dummy(1.0, 1, rand(3,3))

tmp["dummy"] # now the struct itself instead of a dictionary
struct2dict(tmp["dummy"]) # Now a dict where the fields of the structs are now the keys of the dict

# To get desired behavior when calling collect_results we do the following
tabulated_results = collect_results("./", load_function_wrapper = (d) -> struct2dict(d["dummy"]))
```

All 587 tests passed. 

I did try to implement something along the lines of the following.

```julia
function collect_results!(filename, folder;
    valid_filetypes=[".bson", "jld", ".jld2"],
    subfolders=false,
    rpath=nothing,
    verbose=true,
    update=false,
    newfile=false, # keyword only for defining collect_results without !
    rinclude=[r""],
    rexclude=[r"^\b$"],
    load_function_wrapper=(args...; kwargs...) -> wload(args...; kwargs...),
    kwargs...)
```
But I had difficulties getting this to work with `to_data_row` for the specialized JLD2 case. In that case it doesn't expect `wload` (as far as I could tell, I'm not an expert on JLD2 though) so that as a default doesn't work. But I am happy to refactor this PR if there is a better or more idiomatic way to achieve this.

### DIFF
--- a/src/result_collection.jl
+++ b/src/result_collection.jl
@@ -50,7 +50,7 @@ See also [`collect_results`](@ref).
 * `black_list = [:gitcommit, :gitpatch, :script]`: List of keys not to include from result-file.
 * `special_list = []`: List of additional (derived) key-value pairs
   to put in `df` as explained below.
-* `load_function_wrapper = nothing`: Wrapper around calls to loading the data. This is useful in case you need to do some conversion to your data before `collect_results` processes it. For example, if you are loading a struct, it will be stored as a one-element dictionary on disc. When processing it with `collect_results` you might want to convert the struct to a dict so that the fields of the struct become the elements of the dictionary. One way to do this would be to pass `load_function_wrapper = (d) -> struct2dict(d["key_name"])`. Note, this assumes the session you are loading the struct into is aware of the struct type and that the struct definition has not changed. 
+* `load_function = (filename) -> wload(filename)`: function for loading data from file. This is useful in the event you have data saved as a struct. When loaded from file it will be as a one-element `Dict` which is not what you want passed to the dataframe. Instead, you'd rather have the fields of the struct to be available as columns of the dataframe. In that case you can use this function to ensure the struct is converted to a dict before being processed by `collect_results!`. For example, `load_function = (filename) -> struct2dict(wload(filename)["my_key"])`.
 
 `special_list` is a `Vector` where each entry
 is a derived quantity to be included in `df`. There are two types of entries.
@@ -92,7 +92,7 @@ function collect_results!(filename, folder;
     newfile=false, # keyword only for defining collect_results without !
     rinclude=[r""],
     rexclude=[r"^\b$"],
-    load_function_wrapper=nothing,
+    load_function=(filename) -> wload(filename),
     kwargs...)
 
     @assert all(eltype(r) <: Regex for r in (rinclude, rexclude)) "Elements of `rinclude` and `rexclude` must be Regex expressions."
@@ -103,11 +103,7 @@ function collect_results!(filename, folder;
         mtimes = Dict{String,Float64}()
     else
         verbose && @info "Loading existing result collection..."
-        if isnothing(load_function_wrapper)
-            data = wload(filename)
-        else
-            data = load_function_wrapper(filename)
-        end
+        data = load_function(filename)
         df = data["df"]
         # Check if we have pre-recorded mtimes (if not this could be because of an old results database).
         if "mtime" ∈ keys(data)
@@ -177,7 +173,7 @@ function collect_results!(filename, folder;
         mtimes[file] = mtime_file
 
         fpath = rpath === nothing ? file : joinpath(rpath, file)
-        df_new = to_data_row(FileIO.query(fpath); load_function_wrapper=load_function_wrapper, kwargs...)
+        df_new = to_data_row(FileIO.query(fpath); load_function=load_function, kwargs...)
         #add filename
         df_new[!, :path] .= file
         if replace_entry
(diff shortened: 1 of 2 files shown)


src/result_collection.jl
```diff
@@ -100,7 +103,11 @@ function collect_results!(filename, folder;
         mtimes = Dict{String,Float64}()
     else
         verbose && @info "Loading existing result collection..."
-        data = wload(filename)
+        if isnothing(load_function_wrapper)
+            data = wload(filename)
+        else
+            data = load_function_wrapper(filename)
```
Datseris, line 109: Why are we using `nothing` here? Why don't we have the keyword itself be the function? Make `load_function = wload` and then use `load_function` throughout the source code.

src/result_collection.jl
```diff
@@ -50,6 +50,7 @@ See also [`collect_results`](@ref).
 * `black_list = [:gitcommit, :gitpatch, :script]`: List of keys not to include from result-file.
 * `special_list = []`: List of additional (derived) key-value pairs
   to put in `df` as explained below.
+* `load_function_wrapper = nothing`: Wrapper around calls to loading the data. This is useful in case you need to do some conversion to your data before `collect_results` processes it. For example, if you are loading a struct, it will be stored as a one-element dictionary on disc. When processing it with `collect_results` you might want to convert the struct to a dict so that the fields of the struct become the elements of the dictionary. One way to do this would be to pass `load_function_wrapper = (d) -> struct2dict(d["key_name"])`. Note, this assumes the session you are loading the struct into is aware of the struct type and that the struct definition has not changed.
```
Datseris, line 53: I think `load_function_wrapper` is unecessarily complicated. There is no reason to talk about wrappers here. We can keep it simple and say the truth: this is just the loading function. We can call it `load_function`.

Datseris (changes requested):
All good, besides the minor changes for the source code, you should also add a test in the test suite using exactly the situation you describe in the docstring as well.

src/result_collection.jl
NuclearPowerNerd: If you take a look at the pull request I mention why underneath the code snippet. Probably should have included it farther up in the pull request. 

> But I had difficulties getting this to work with to_data_row for the specialized JLD2 case. In that case it doesn't expect wload (as far as I could tell, I'm not an expert on JLD2 though) so that as a default doesn't work. But I am happy to refactor this PR if there is a better or more idiomatic way to achieve this.

Datseris: Sorry, but I don't understand the problem. Please explain it again. The way I see the code right now, if `load_function` is nothing, it becomes `wload`. Otherwise, it becomes a user-provided function. So... `load_function` can be by default `wload`. That's what it becomes anyways. So I am overall confused because in the current PR `wload` is the default but you say it doesn't work as the default.

The quote you quoted: what does "this" mean in its first sentence?

NuclearPowerNerd: > Sorry, but I don't understand the problem. Please explain it again. The way I see the code right now, if `load_function` is nothing, it becomes `wload`. Otherwise, it becomes a user-provided function. So... `load_function` can be by default `wload`. That's what it becomes anyways. So I am overall confused because in the current PR `wload` is the default but you say it doesn't work as the default.
> 
> The quote you quoted: what does "this" mean in its first sentence?

Yeah sorry for the confusion, I probably made this more complicated than it needed to be. I've pushed some new changes based on your advice. Thanks for your patience. Please take a look and let me know what you think.

NuclearPowerNerd:
> All good, besides the minor changes for the source code, you should also add a test in the test suite using exactly the situation you describe in the docstring as well.

Thanks - I've updated the pull request. Also just realized my autoformat in vscode made tons of changes to `update_results_tests.jl`. That probably makes it harder to see my actual changes, which begin on line 73 of that file.

src/result_collection.jl
```diff
@@ -50,6 +50,7 @@ See also [`collect_results`](@ref).
 * `black_list = [:gitcommit, :gitpatch, :script]`: List of keys not to include from result-file.
 * `special_list = []`: List of additional (derived) key-value pairs
   to put in `df` as explained below.
+* `load_function = (filename) -> wload(filename)`: function for loading data from file. This is useful in the event you have data saved as a struct. When loaded from file it will be as a one-element `Dict` which is not what you want passed to the dataframe. Instead, you'd rather have the fields of the struct to be available as columns of the dataframe. In that case you can use this function to ensure the struct is converted to a dict before being processed by `collect_results!`. For example, `load_function = (filename) -> struct2dict(wload(filename)["my_key"])`.
```
Datseris, line 53:
```suggestion
* `load_function = wload`: function for loading data from file. This is useful in the event you have data saved as a struct. When loaded from file it will be as a one-element `Dict` which is not what you want passed to the dataframe. Instead, you'd rather have the fields of the struct to be available as columns of the dataframe. In that case you can use this function to ensure the struct is converted to a dict before being processed by `collect_results!`. For example, `load_function = (filename) -> struct2dict(wload(filename)["my_key"])`.
```

Datseris: `wload` is already a function. Wrapping it in an anonymous function has no benefit.

src/result_collection.jl
```diff
@@ -72,24 +73,26 @@ and then define
...
-    rexclude = [r"^\b$"],
+    valid_filetypes=[".bson", "jld", ".jld2"],
+    subfolders=false,
+    rpath=nothing,
+    verbose=true,
+    update=false,
+    newfile=false, # keyword only for defining collect_results without !
+    rinclude=[r""],
+    rexclude=[r"^\b$"],
+    load_function=(filename) -> wload(filename),
```
Datseris, line 95:
```suggestion
    load_function = wload,
```

Datseris (changes requested):
Sorry for late. Don't worry about windows failures we have a problem that we don't know how to solve and CI fails on Windows.

There is a small change necessary: there is no reason to use an anonumous wrapper.

There is also a larger change necessary: please revert all formatting changes. Probably simpler to start a new branch/pr or reset this one to its original state and then do the couple of lines changes necessary.

NuclearPowerNerd:
> Sorry for late. Don't worry about windows failures we have a problem that we don't know how to solve and CI fails on Windows.
> 
> There is a small change necessary: there is no reason to use an anonumous wrapper.
> 
> There is also a larger change necessary: please revert all formatting changes. Probably simpler to start a new branch/pr or reset this one to its original state and then do the couple of lines changes necessary.

Thanks - I've updated the pull request per your comments.

NuclearPowerNerd:
So I've prepared a new branch without the whitespace changes. Would it just be easier to close this PR and open a new one with the cleaned up branch?

Datseris:
sorry for the late reply!!! Yes, a new PR is easiest!

Issue #399, opened 2023-10-17
Move to Julia 1.9, use extensions remove Requires for `collect_results` and Dataframes


Datseris:
The title says it all, currently we use Requires for `collect_results`. We should not do that as Julia natively supports conditional dependencies via the extensions. See here for how to do it: https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)

denfc:
>We should not do that 

Via `PkgDependency`, I see that DrWatson v2.15.0. still uses Requires.  Is an update on the way or should I learn to implement conditional dependencies?  If the latter, could you post the code you use? 

Thanks!

-- dfc

Datseris:
this has the code/instructions for how to do it.

PR #423, opened 2024-08-20
Issue #399

franzi448:
Added conditional loading for extensions for 'collect_results'.


Datseris:
Thanks for the effort @franzi448 , but I think if we do this we should formally move to 1.9 Julia version and remove `Requires` completely from the source code. Given that 1.10 becomes the new LTS version, I am feeing comfortable to increment the Julia version in Project.toml.

franzi448:
Alright, so should I remove the 'Requires' pkg from the project.toml and main code or should I not proceed with the PR?

Datseris:
Sorry, I wasn't clear: remove `Requires` from the source code, and bump Julia version to 1.9!

PR #435, opened 2025-07-22
Avoid deadlock when reading stdout and stderr from commands

### LINKED ISSUE #300: gitpatch hangs sometimes for dirty repos
Sometimes, when my working directory is dirty, `gitpatch` will hang.
After `Ctrl-C`ing out, everything still works and simulations resume as desired.

```
┌ Warning: The Git repository ('/[...].jl') is dirty! Appending _dirty to the commit ID.
└ @ DrWatson /[...]/.julia/packages/DrWatson/YWCBx/src/saving_tools.jl:65
^C┌ Warning: `gitpatch` failed with error  InterruptException() and , returning `nothing` instead.
└ @ DrWatson /[...]/.julia/packages/DrWatson/YWCBx/src/saving_tools.jl:144
```

Unfortunately, I don't know, when exactly this happens. Usually, for dirty repos it just prints this message and resumes. Maybe this can happen, if I create a new commit or create a new branch while the simulations are running. Usually, as soon as this behavior shows up, the future `Appending _dirty to the commit ID.` will also hang.

danielmatz:
The `gitpatch` function sometimes hangs for large git diffs. This seems to be happening because the `read_stdout_stderr` function is writing to the `out` and `err` pipes, but no task is reading from them, so they fill up, and the process locks up waiting to write more data.

This change spawns separate tasks to read from the `out` and `err` pipes, thus avoiding the deadlock.

The Julia compat for this package is set to 1.0, so I opted to use `@async` instead of `Threads.@spawn`.

Fixes #300

### DIFF
--- a/src/saving_tools.jl
+++ b/src/saving_tools.jl
@@ -109,6 +109,8 @@ Return a `NamedTuple` with the fields `exception`, `out` and `err`.
 function read_stdout_stderr(cmd::Cmd)
     out = Pipe()
     err = Pipe()
+    out_string = @async read(out, String)
+    err_string = @async read(err, String)
     exception = nothing
     try
         run(pipeline(cmd,stderr=err, stdout=out), wait=true)
@@ -117,7 +119,7 @@ function read_stdout_stderr(cmd::Cmd)
     end
     close(out.in)
     close(err.in)
-    return (exception = exception, out=read(out,String), err=read(err,String))
+    return (exception = exception, out = fetch(out_string), err = fetch(err_string))
 end
 
 """


Datseris (commented):
Thank you very much, this seems to be solving a really old issue! I am wondering how you found this!

In any case, is this possible to test somehow? I am not sure... If you can't think of a test it's okay. 

You should increase Project.toml patch version.

Datseris:
Test failure is unrelated.

danielmatz:
> In any case, is this possible to test somehow? I am not sure... If you can't think of a test it's okay.

Is this not already being tested? I see in `test/stools_tests.jl` that you have a temporary git repo with uncommitted changes, and you are testing saving the patch.

Datseris (approved):

PR #437, opened 2025-07-23
Add `SubString` to `default_allowed`

### LINKED ISSUE #436: May want to add `SubString` to `savename`'s `allowedtypes` (ergo in `default_allowed`)?
Thanks for writing and maintaining this library! It's been very useful to me! 🙂 

I've recently tripped several times over the fact that `SubString` is not one of the default `allowedtypes`. `SubString`s look the same as `String`s in the REPL on first glance:

```julia
julia> "uiae"
"uiae"
julia> typeof("uiae")
String
julia> split("uiae", "i")[1]
"u"
julia> typeof(split("uiae", "i")[1])
SubString{String}
```

`SubString`s are created often by string operations (e.g. `split`). If `param` is a `SubString` (for whatever reason—e.g. because it is parsed somewhere), then 

```julia
savename("something", (;param), "txt")
```

silently ignores `param` despite it being given explicitly. This can lead to overwriting earlier results.

dpaetzel:
Fixes #436.

I kept the change set small; it would have been possible to also adjust the examples (in particular, [the real world example](https://github.com/JuliaDynamics/DrWatson.jl/blob/main/docs/src/real_world.md)), but since those did not use the most up-to-date definition of `default_allowed` anyway, I abstained from doing so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 2.19.0
+
+- Add `SubString` to `default_allowed` so that `savename(prefix, (; a=a),
+  suffix)` works as expected if `a` was produced by `String` operations such as
+  `split`.
+
 # 2.18.0
 
 - `tag!` and parent functions allow keyword `warn`.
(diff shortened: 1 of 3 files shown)


Datseris:
Thank you. Can you please bump minor version in project.toml and add a changelog entry stating this?

PR #440, opened 2025-09-13
Add macro and functions for named tuple updates

aquaresima:
Hello, 

This file introduces a macro for updating named tuples and a deep merge function to handle nested updates. It also includes a pretty print function for displaying named tuples. 

I wrote this macro for my personal use, but because I use DrWatson in all my scientific projects, and I believe that this is a valuable tool for anyone working with nested tuples, I think it would be nice to have it here!

The idea is to use it to update experiment configurations within loops.

It works like this:

```julia

## Example usage:
base = (a=5, b=(d=6, f=7), c=(to="be", or="not to be"))

new_config = @update base begin
    b.e = (new="field", m=66) # Adding a new nested field
    b.g.nested = (;new ="field") # Adding a deeper nested field
    b.a = "This is changed" # Changing 
    c = "to be changed" # Changing an existing field
end

println("Base configuration:")
pretty_nt_print(base)

println("\nUpdated configuration:")
pretty_nt_print(new_config)
```

```
Base configuration:

{
    a := 5
    b := {
      d := 6
      f := 7
     }
    c := {
      to := be
      or := not to be
     }
   }


Updated configuration:

{
    a := 5
    b := {
      d := 6
      f := 7
      e := {
        new := field
        m := 66
       }
      g := {
        nested := {
          new := field
         }
       }
      a := This is changed
     }
    c := to be changed
   }
```

There is also an in-place `@update!` version

```julia
@update! base begin
    b.e = (l=65, m=66) # Adding a new nested field
    b.g.first = "Nested" # Adding a deeper nested field
    b.a = "This is changed" # Changing 
    c = "to be changed" # Changing an existing field
end

# @assert base == new_config

base = @update base a = "inline" 
@update! base b = "inline with !" 
```

### DIFF
--- a/src/named_tuple_update.jl
+++ b/src/named_tuple_update.jl
@@ -1,0 +1,139 @@
+#Update macro and functions for handling named tuples
+
+# Simple update macro for handling multiple updates in a block
+macro update(base, update_expr)
+    # Verify if the expr is a block or a line
+    if update_expr.head == :block
+        # if a block, extract the expressions
+        updates = update_expr.args
+
+        # Start with the base configuration
+        # The :($(esc(base))) is used to ensure the base is evaluated in the correct context (the macro's context)
+        current_config = :($(esc(base)))
+
+        # Process each update expression in the block
+        for update_expr in updates
+            isa(update_expr, LineNumberNode) && continue  # Ensure it's an expression
+
+            # Extract the left-hand side and right-hand side, the left-hand side is the field to update, the right hand side is the new value
+            lhs, rhs = update_expr.args
+
+            # Escape the value to ensure it's evaluated in the correct context
+            value = :($(esc(rhs)))
+
+            # Assert the left-hand side has the correct structure
+            # if isa(lhs, Symbol)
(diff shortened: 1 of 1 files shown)


aquaresima:
Before merging, if you are interested anyway, one should decide whether to maintain or not the warning for the creation/deletion of fields.

Currently, creating a nested field that was not previously there or replacing a field with a value using a named tuple generates a warning. However, I am uncertain about the practicality of this feature within the library's broader context.

Because the user is modifying the named tuples "in batch", this framework is as useful as prone to errors. Thus, the warnings are convenient to avoid making the most common mistake: 
- You think that the field you want to update is `a.b.c` while it is `a.c`

Tho, this can be done on purpose, and in that case, having the warning at run-time can be pretty annoying.

Datseris:
Hi there, thanks for the contribution! I agree that this is generally useful functionality, however it is probably a better fit for https://github.com/JeffreySarnoff/NamedTupleTools.jl . There is some functionality overlap there as well. Can you try to add it as a contribution there?

